PR #9, opened 2015-06-11
Julia v0.4 compatibility

alyst:
There are `Tuple` and `Dict` changes in Julia v0.4 that are not compatible with `BlackBoxOptim`.
The PR updates `Tuple` and `Dict` syntax and also fixes a few other API deprecations.
It uses `Compat` package for backward-compatibility, so it should be compatible with v0.3.x, but I have not tested it.
It also makes a few tiny enhancements and simplifications that I observed during the code migration (int conversions, the usage of abstract types in types and function declarations, replacing Strings by Symbols as dictionary keys).

PR #10, opened 2015-06-11
Column-wise storage of individuals in the population matrix

alyst:
As the arrays are stored in column-wise manner in Julia (i.e individual column occupies continuous memory block), it should be more efficient to store the individuals as columns in the population matrix.
This is what the PR addresses. Column-wise storage also makes the things more straightforward as now the individuals are 1D vectors (not 2D row-vectors as before).

I see some indications that it improves the performance, but I don't have a proper benchmark. The problems in the test suite are low-dimensional, the real test should be some high-dimensional problem with a fast-computing functional.

I've also made some refactoring of the `Archive` class and replaced tuples by immutables, so that it is easier to read the code. Plus, replaced `Union{Nothing,Vector}` with the `Nullable{Vector}` as it is exactly what `Nullable` is required for.

It's based on #9 branch.


robertfeldt:
Thanks. Yes, I wrote this early in my Julia career... 

I'm not sure this is a very important one since we might need to move to non-matrix based representations over time anyway. To evolve/optimize non-vector representations etc. But I'm willing to consider it if all tests pass and I see no adverse affects in projects were we are already using BBO package. But lets close the v0.4 compat first.

alyst:
The API might evolve to allow different population implementations, but I guess the matrix-based one should stay as it's the most efficient and straightforward.

robertfeldt:
Most of this looks good and tests pass on my end. However, I'm a bit concerned about line 99 in differential_evolution.jl. Hard when you mix in refactorings with possible behavior change. Reasoning for that change?

robertfeldt:
Have you verified that rand_individuals does not need to change despite the changed col/row order?

alyst:
`index=psize` would be mapped into `psize`, and then for `index=psize+1` `mod(index,psize)+1 = 2`, so in the first "wrapping" `1` is skipped. The patch fixes that to ensure the uniformity of the indices distributions.

alyst:
Yes, I was also a little bit confused with `rand_individuals()`, but the current code looks correct (you can try e.g. `broadcast( +, [0, 10, 20], broadcast( *, [1,0.1,0.01], rand(3,5) ) )`).
Actually, the older code for `rand_individuals()` was also "column-wise", but, if I'm not mistaken, it was only used by the non population-based methods. The differential evolution is using latin cube-based generation, which had row-wise individuals.
So now the parameter set representation should be more homogeneous across the whole range of methods.

robertfeldt:
Yes, it's actually a bug fix. Great, big thanks. I'll merge and then try to add a few tests to try to capture this/similar problem(s) in the future.

PR #12, opened 2015-06-13
Update DE and ADE to use GeneticOperator approach

alyst:
It's me again. The package already had the code for genetic operators (mutation, crossover etc) abstraction, in this PR I convert DE and ADE algorithms to actually use it. There are some key advantages of operators:
 * operator-specific parameters could be encapsulated in the operator's class; that is not possible when using functions
 * specifying genetic operators as the parameters of an algorithm class facilitates compile-time optimizations of the algorithm code

I also introduced DiffEvoParameters class that allows unifying code for ADE and DE.

Finally, the crossover operators were optimized to only calculate the new values for the dimensions that are actually crossed-over.

### DIFF
--- a/src/BlackBoxOptim.jl
+++ b/src/BlackBoxOptim.jl
@@ -1,5 +1,7 @@
 module BlackBoxOptim
 
+using Distributions, Compat
+
 export  Optimizer, PopulationOptimizer,
         bboptimize, compare_optimizers,
 
@@ -44,8 +46,6 @@ export  Optimizer, PopulationOptimizer,
 
 abstract Optimizer
 
-using Compat
-
 module Utils
   include("utilities/latin_hypercube_sampling.jl")
   include("utilities/assign_ranks.jl")
@@ -55,6 +55,10 @@ include("search_space.jl")
 include("parameters.jl")
 include("fitness.jl")
 include("population.jl")
+
+# Genetic Operators
+include("genetic_operators/genetic_operator.jl")
+
 include("frequency_adaptation.jl")
 include("archive.jl")
 
@@ -130,9 +134,6 @@ include("direct_search_with_probabilistic_descent.jl")
 
 # End-user/interface functions
 include("bboptimize.jl")
-
-# Genetic Operators
-include("genetic_operators/genetic_operator.jl")
 
 # Fitness
 include("fitness/fitness_types.jl")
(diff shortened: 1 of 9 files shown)


robertfeldt:
Thanks a lot, this looks good. I started on the new operator design some time back but have had no time this spring to get it in there. Also nice cleanup and refactorings.

Before I merge this I want to create a small script we can use to compare old versions to new ones on a large set of optimization problems. This way we can ensure we do not degrade performance as we make changes that might affect performance. I will check it in later today and then test your new code with it.

robertfeldt:
Sorry this took some time. I need to do a comparison script and run extensive comparisons. On the 18 problems and 4 DE implementation (=72 different settings to compare) I included in the comparison the new implementation was statistically significantly better in 24 of 29 runs with a difference. And in the other five results were:

```
"Problem","NumDims","Method","OldMedianFitness","Order","NewMedianFitness","Pvalue","Level","StatSignDiff"
Rosenbrock",30,"de_rand_1_bin_radiuslimited",10.280394067632201,"<",12.563219593612112,0.00038222316468955404,0.01,"Yes"
"Rosenbrock",30,"de_rand_1_bin_radiuslimited",10.280394067632201,"<",13.323925765971003,6.264251916504846e-5,0.01,"Yes"
"Schwefel2.22",5,"adaptive_de_rand_1_bin_radiuslimited",1.3774708862931864e-8,"<",1.9042861898492163e-8,0.007601641475220176,0.01,"Yes"
"Schwefel2.22",30,"adaptive_de_rand_1_bin_radiuslimited",1.1855387823729556e-8,"<",2.3973681367645136e-8,2.695471360038315e-5,0.01,"Yes"
"Schwefel2.21",10,"adaptive_de_rand_1_bin_radiuslimited",0.010604498603295709,"<",0.013919701853145153,0.00927263490774628,0.01,"Yes"
```

so I'll go ahead and merge the pull request. Since my vacation started now I hope to have more time to work on BBO and accept any future requests quicker. If you are looking into performance there is a big waste now in that we always re-evaluate old individuals even if the fitness function is not noisy. This is why there is a double number of fitness evaluations compared to the number of steps, for most optimization methods. Should be quite easy to fix a parameter for that.

alyst:
Great, thank you! Meanwhile I have accumulated some other changes, I'll try to submit the PRs soon.

I was also thinking about caching the fitness, but I have not figured out for myself what would be the best strategy. There's ```Evaluator``` (and ```Archive```) and ```Population```. I would say that naturally this cached fitness belongs to [subclass of] ```Population``` (it's easier to keep it in sync with the changes in population), but then there would be an overlap of the functionality between ```Population``` and ```Evaluator```/```Archive```.

robertfeldt:
Yes, I never went ahead with this fix since it was not clear how to design it. IMHO it does not belong with Archive. I think we should add an option to Evaluator that can take fitness objects as inputs and then (depending on the Evaluator) it might or might not decide/need to re-evaluate. Then it would be natural to keep the fitness values in an array in the Population object imho. Optimizers that are not population-based can still use a Population of size 1 (not sure what to call that though) and thus have the same interface to get fitness etc.

I think we should also soon consider using pmap in Evaluators to speed up evaluation for people that have long-running fitness funcs.

multidis:
`pmap` for objective evaluation would help indeed

PR #13, opened 2015-06-29
Speedup fitness_is_within_ftol()

alyst:
I was doing some profiling using ```@profile bboptimize()``` and then looking at the results with ```ProfileView.jl```. It was quite surprising to find that the program spent 5% of its time in ```fitness_is_within_ftol()```, which is a fairly simple function. The reason was ```try/catch```, which is very costly.
The PR changes the TopListArchive so that ```try/catch``` is not required anymore.

### DIFF
--- a/src/archive.jl
+++ b/src/archive.jl
@@ -46,16 +46,16 @@ type TopListArchive <: Archive
   end
 end
 
-capacity(a::Archive) = a.capacity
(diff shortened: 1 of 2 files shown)


robertfeldt:
Interesting since I have done no profiling or perf optimization so far. Good one.

PR #14, opened 2015-06-29
Misc cosmetic changes

alyst:
This is just some small cosmetic changes + better v0.4 compatibility. Performance should not be affected.
The biggest change is the split of `setup_problem()` into 3 methods -- by the type of a `problem` argument. It makes Julia do the type-matching job for us and discards big outer ```if/else```.


robertfeldt:
Looks good. Fails on nightly julia build but seems to be a problem with the build rather than your changes. Everything works nicely on my side on 0.3.9 and a few days old 0.4. So should be fine.

alyst:
@robertfeldt Thanks! I guess the nightlies failure would be resolved, when ```Distributions.jl``` would be synced with ```@math_const``` changes.

PR #15, opened 2015-07-03
Embed and Select operators

alyst:
The PR adds ```EmbeddingOperator``` (projection into the search space, but I called it "embedding", because it could be more sophisticated that just projection, also not necessarily onto the bounds of the search space) and ```IndividualsSelector``` operator (picking the individuals from the population).
With this the type parameterization of ```DiffEvoOpt``` should be complete.
As all the required properties are now the explicit fields of the appropriate types, the ```options``` dictionary is removed from the ```DiffEvoOpt```. I think the use of dictionaries to specify the options should be minimized as it's harder to keep track and synchronize the required parameters, and the problems are only revealed at runtime.


robertfeldt:
Looks good to me. Will benchmark and then merge.

I agree in general with avoinding options dictionaries but users probably want a simple interface at the "top level". So maybe a good trade-off is that we use options dicts for the bboptimize and the "shortand" ctr functions (adaptive_de_bin_1 etc) but use normal args to the actual Optimizer types.

alyst:
@robertfeldt Yes, that's the idea -- to have a simple "quick and dirty" user interface like ```bboptimize()```, but be strict in the internal code. I think it would be also benefitial for advanced users who want more control over the optimization process (their own genetic operators etc).

robertfeldt:
Ok, benchmarking seems fine:
```
Num problems (any p-value) = 92
  Num where new implementation is better = 41
  Num where new implementation is worse = 51
Num problems (with p-values < 0.05) = 0
  Num where new implementation is better = 0
  Num where new implementation is worse = 0
Num problems (with p-values < 0.01) = 0
  Num where new implementation is better = 0
  Num where new implementation is worse = 0
```

alyst:
@robertfeldt Thanks! Are the benchmarking comparisons based on number of steps or seconds?

robertfeldt:
Number of function evals currently. So the DE's are at a disadvantage compared to compass search et al. We will fix that with the new candidate type and smarter fitness eval.

alyst:
Actually, I already have an early version of storing the fitness in the population. I will create a pull request, when it would be polished a little bit.

PR #16, opened 2015-07-03
MutationClock and MutationMixture

alyst:
The PR makes ```MutationClock``` working by adding ```GibbsMutationOperator``` types (mutates one specific dimension at a time) and adds ```MutationMixture``` that randomly applies the mutation from a list according to the weights.

It requires #15 to be merged first, so when and if that would be done, I'll rebase this PR.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,5 @@
 julia 0.3-
 FactCheck
+StatsBase
 Distributions
 Compat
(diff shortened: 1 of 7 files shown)


robertfeldt:
Ok, please rebase so I can check. Thanks.

alyst:
done, tests pass

robertfeldt:
Sorry but this is not how the MutationClock is supposed to work imho. It should not be guaranteed that at least one dim is mutated; that should depend on the mutation probability. If mutation probability is very low there can be many calls to the mutation clock operator in which nothing happens.

alyst:
Oh, thanks for noticing. I've reverted that.

PR #17, opened 2015-07-03
Candidate type

alyst:
This introduces ```Candidate``` type that gets passed by ```ask()```/```tell!()``` calls. It replaces the tuple-based approach.
It would allow filling ```fitness``` field already in ```ask() ```, where we have access to the population (that might cache the fitness of individuals in future), so we can avoid reevaluating the fitness in ```rank_by_fitness!()```.

Also, in preparation to ```Candidate``` introduction, I've made some memory/performance optimizations in  NES.

This requires #16 to be merged first.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,5 @@
 julia 0.3-
 FactCheck
+StatsBase
 Distributions
 Compat
--- a/src/BlackBoxOptim.jl
+++ b/src/BlackBoxOptim.jl
@@ -1,6 +1,6 @@
 module BlackBoxOptim
 
-using Distributions, Compat
+using Distributions, StatsBase, Compat
 
 export  Optimizer, PopulationOptimizer,
         bboptimize, compare_optimizers,
@@ -44,7 +44,7 @@ export  Optimizer, PopulationOptimizer,
 
         # Genetic operators
         GeneticOperator, MutationOperator, CrossoverOperator, EmbeddingOperator,
-        NoMutation, MutationClock, MutationMixture, GibbsMutationOperator, SimpleGibbsMutation,
+        NoMutation, MutationClock, GibbsMutationOperator, SimpleGibbsMutation, MutationMixture,
         RandomBound,
         SimpleSelector, RadiusLimitedSelector,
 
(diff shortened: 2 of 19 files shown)


robertfeldt:
Overall I think a Candidate{F} type is the way to go but it seems rather strange to make them just for the sake of evaluation. Isn't it more natural that a population is always a Vector of Candidate's? And that Evauators's get not the params but the candidate's and then only re-evaluates the fitness if they are a noisy evaluator or if, say, a version number has changed (there are some fitness functions that gradually learn the max and min values over all evaluations, for example, so they need version numbers to recalc from raw fitness values)?

alyst:
Yes, the concept is that Candidate is some individual, not necessarily from the population, that we evaluate and then either accept into the population or throw away.
For DE and similar algorithms making the population a vector of Candidates makes sense and would not affect the performance (although we don't need the ```Candidate.index``` field within the population). Unfortunately, my knowledge of all available global optimization approaches is rather superficial, so I don't know whether vector-of-Candidates would be optimal for all the algorithms. In xNES it seems all the individuals are changed using matrix algebra, so Matrix could still be an optimal individuals storage there (the ranking, though, as implemented in this PR is inefficient for NES -- we don't need to copy the matrix column into another vector as we are not going to modify it externally, it's enough just to pass the column index. That could be addressed by an alternative implementation e.g. ```CandidateRef{F}``` that only stores the index of the individual).
Storing the fitness was not the part of this PR as it would affect too many places in the code, but that would be exactly the idea -- ```rank_by_fitness!()``` would skip evaluation if fitness is already there.

robertfeldt:
I think for > 90% of opt algs we should be fine with a vector of candidates as the population but you are right that there are some were we might need to use a matrix, yes. Ok, sounds good then. I'll merge after benchmarking and after other PR.

robertfeldt:
Need to be rebased, right?

robertfeldt:
Actually I think using the Poisson for the MutationClock is not correct since it gives the number of events in a time interval, not the times of the events. We may need to revert this MutationClock change and we should probably add a test to check even distribution over param indices.

alyst:
@robertfeldt Sorry for Poisson. That was mean to be Exponential, some aberration of my mind. Thanks for merging!

robertfeldt:
No problem. Thanks for your work on BBO. Seeing that people care and actually help out with this project motivates me to also work more on it and integrate some of our recent research into this. Once we have the "new" design of evaluation/candidates/ask/tell in there I can start adding more of the algorithms that are already implemented as spikes. I'm also working on integrating my multi-objective code from other projects as well as adding Escher.jl interface to running opts and compare script.

PR #18, opened 2015-07-09
Fitness-storing Population & friends

alyst:
The major goal of the patch is ```FitPopulation{F}``` type that allows storing fitness values within the population and ```rank_by_fitness!()``` that skips the candidates with the defined fitness, which should double the performance of DE.

The PR is quite invasive (mostly motivated by the need to have fitness value type known by the population and fitness schema known by the optimization problem) and touches a few core classes:
  * ```FitnessScheme``` is parameterized by fitness value type. The available schemes are ```ScalarFitness{true/false}``` and ```VectorFitness{true/false}``` (this one untested)
  * ```FitnessScheme``` is now accessible from ```OptimizationProblem``` as well as the ```SearchSpace```
  * ```FixedDimProblem``` renamed into ```FunctionBasedProblem``` and ```AnyDimProblem``` into ```FunctionBasedProbelmFamily```. The latter is also not an ```OptimizationProblem``` anymore  --- that's because all the problems the optimizers could solve [currently] are finite dimensional
  * ```FunctionBasedProblem``` is subclassed into ```Bounded``` (optimum is known) and ```Unbounded``` -- the required subtype is determined automatically
  * ```AskTellOptimizer``` abstract type
  * ```Optimizer``` constructors now have access to the ```OptimizationProblem```, therefore passing ```search_space``` through dictionary no longer needed
  * to reduce the GC overhead ```FitPopulation``` maintains a pool of candidates -- no memory allocation in ```ask()/tell!()```, just copying.

### DIFF
--- a/examples/benchmarking/compare_optimizers.jl
+++ b/examples/benchmarking/compare_optimizers.jl
@@ -184,10 +184,10 @@ increase_runs_per_problem(problemname, numdims) = begin
   RunsPerProblem[k] = get(RunsPerProblem, k, 0) + 1
 end
 
-function fitness_for_opt(problem, numDimensions, populationSize, numFuncEvals,
+function fitness_for_opt(family::FunctionBasedProblemFamily, numDimensions, populationSize, numFuncEvals,
     method, showtrace = true)
 
-  problem = BlackBoxOptim.as_fixed_dim_problem(problem, numDimensions)
+  problem = BlackBoxOptim.fixed_dim_problem(family, numDimensions)
 
   best, fitness = bboptimize(problem; method = method, parameters = {
     :NumDimensions => numDimensions,
--- a/examples/benchmarking/compare_optimizers.jl
+++ b/examples/benchmarking/compare_optimizers.jl
@@ -200,7 +200,7 @@ function fitness_for_opt(family::FunctionBasedProblemFamily, numDimensions, popu
 end
 
 function latest_git_id()
-  strip(readall(`git log --format="%H" -n 1`))
+  strip(readall(`git -C "$(Pkg.dir("BlackBoxOptim"))" log --format="%H" -n 1`))
 end
 
 # Test an optimizer on multiple problems and dimensions, return results in dict.
--- a/src/BlackBoxOptim.jl
+++ b/src/BlackBoxOptim.jl
@@ -30,7 +30,7 @@ export  Optimizer, AskTellOptimizer, SteppingOptimizer, PopulationOptimizer,
         Problems,
         OptimizationProblem, FunctionBasedProblem,
         name, fitness_scheme, search_space, numdims, opt_value,
-        fitness_is_within_ftol, objfunc,
+        fitness_is_within_ftol, objfunc, fitness,
 
         # Problem factory/family
         FunctionBasedProblemFamily, MinimizationProblemFamily,
@@ -50,7 +50,7 @@ export  Optimizer, AskTellOptimizer, SteppingOptimizer, PopulationOptimizer,
         rand_individual, rand_individuals, isinspace, rand_individuals_lhs,
 
         # Population
-        FloatVectorPopulation,
+        FitPopulation,
         popsize,
 
         # Genetic operators
@@ -75,6 +75,7 @@ abstract AskTellOptimizer <: Optimizer
 # population-based optimizers
 abstract PopulationOptimizer <: AskTellOptimizer
 population(popopt::PopulationOptimizer) = popopt.population
+popsize(popopt::PopulationOptimizer) = popsize(population(popopt))
 
 module Utils
   include("utilities/latin_hypercube_sampling.jl")
@@ -84,7 +85,6 @@ end
 include("search_space.jl")
 include("parameters.jl")
 include("fitness.jl")
-include("population.jl")
 
 # Genetic Operators
 include("genetic_operators/genetic_operator.jl")
@@ -115,6 +115,9 @@ function name(o::Optimizer)
     return s
   end
 end
+
+# Population
+include("population.jl")
 
 # Our design is inspired by the object-oriented, ask-and-tell "optimizer API
 # format" as proposed in:
--- a/src/BlackBoxOptim.jl
+++ b/src/BlackBoxOptim.jl
@@ -2,7 +2,7 @@ module BlackBoxOptim
 
 using Distributions, StatsBase, Compat
 
-export  Optimizer, PopulationOptimizer,
+export  Optimizer, AskTellOptimizer, SteppingOptimizer, PopulationOptimizer,
         bboptimize, compare_optimizers,
 
         DiffEvoOpt, de_rand_1_bin, de_rand_1_bin_radiuslimited,
@@ -50,7 +50,20 @@ export  Optimizer, PopulationOptimizer,
 
         name
 
+# base abstract class for black-box optimization algorithms
 abstract Optimizer
+
+# SteppingOptimizer's do not have an ask and tell interface since they would be
+# complex to implement if forced into that form.
+abstract SteppingOptimizer <: Optimizer
+evaluator(so::SteppingOptimizer) = so.evaluator
+
+# optimizer using ask()/..eval fitness../tell!() sequence at each step
+abstract AskTellOptimizer <: Optimizer
+
+# population-based optimizers
+abstract PopulationOptimizer <: AskTellOptimizer
+population(popopt::PopulationOptimizer) = popopt.population
 
 module Utils
   include("utilities/latin_hypercube_sampling.jl")
@@ -92,10 +105,6 @@ function name(o::Optimizer)
   end
 end
 
-abstract PopulationOptimizer <: Optimizer
-
-population(o::PopulationOptimizer) = o.population # Fallback method if sub-types have not implemented it.
-
 # Our design is inspired by the object-oriented, ask-and-tell "optimizer API
 # format" as proposed in:
 #
@@ -125,8 +134,6 @@ population(o::PopulationOptimizer) = o.population # Fallback method if sub-types
 # is no single optimum. Instead there are many pareto optimal solutions.
 # An archive collects information about the pareto optimal set or some
 # approximation of it. Different archival strategies can be implemented.
-
-has_ask_tell_interface(o::Optimizer) = true # Default is to have an ask and tell interface...
 
 # Different optimization algorithms
 include("random_search.jl")
--- a/src/BlackBoxOptim.jl
+++ b/src/BlackBoxOptim.jl
@@ -27,10 +27,16 @@ export  Optimizer, AskTellOptimizer, SteppingOptimizer, PopulationOptimizer,
         #ProblemEvaluator,
 
         # Problems
-        Problems, FixedDimProblem, is_fixed_dimensional, is_any_dimensional,
-        is_single_objective_problem, is_multi_objective_problem,
-        search_space, eval1, evalall, anydim_problem, as_fixed_dim_problem,
-        fitness_is_within_ftol, save_fitness_history_to_csv_file,
+        Problems,
+        OptimizationProblem, FunctionBasedProblem,
+        name, fitness_scheme, search_space, numdims, opt_value,
+        fitness_is_within_ftol, objfunc,
+
+        # Problem factory/family
+        FunctionBasedProblemFamily, MinimizationProblemFamily,
+        fixed_dim_problem,
+
+        save_fitness_history_to_csv_file,
 
         # Archive
         TopListArchive, best_fitness, add_candidate!, best_candidate,
@@ -86,8 +92,8 @@ include("genetic_operators/genetic_operator.jl")
 include("frequency_adaptation.jl")
 include("archive.jl")
 
-# Problems for testing
 include(joinpath("problems", "all_problems.jl"))
+include(joinpath("problems", "problem_family.jl"))
 
 include("evaluator.jl")
 
@@ -158,5 +164,8 @@ include("bboptimize.jl")
 include("fitness/pareto_dominance.jl")
 include("fitness/epsilon_pareto_dominance.jl")
 include("fitness/epsilon_box_dominance.jl")
+
+# Problems for testing
+include(joinpath("problems", "single_objective.jl"))
 
 end # module BlackBoxOptim
--- a/src/BlackBoxOptim.jl
+++ b/src/BlackBoxOptim.jl
@@ -16,6 +16,12 @@ export  Optimizer, AskTellOptimizer, SteppingOptimizer, PopulationOptimizer,
         Parameters, mergeparam,
 
         # Fitness
+        FitnessScheme,
+        ScalarFitness, ComplexFitness, VectorFitness,
+        fitness_type, numobjectives,
+        is_minimizing, nafitness, isnafitness,
+        hat_compare, is_better, is_worse, same_fitness,
+        vector_fitness_scheme_min, vector_fitness_scheme_max,
 
         # Evaluator
         #ProblemEvaluator,
@@ -37,10 +43,9 @@ export  Optimizer, AskTellOptimizer, SteppingOptimizer, PopulationOptimizer,
         numdims, mins, maxs, deltas, ranges, range_for_dim, diameters,
         rand_individual, rand_individuals, isinspace, rand_individuals_lhs,
 
-        hat_compare, is_better, is_worse, same_fitness,
-        popsize,
-        FloatVectorFitness, float_vector_scheme_min, float_vector_scheme_max,
+        # Population
         FloatVectorPopulation,
+        popsize,
 
         # Genetic operators
         GeneticOperator, MutationOperator, CrossoverOperator, EmbeddingOperator,
@@ -149,7 +154,7 @@ include("direct_search_with_probabilistic_descent.jl")
 include("bboptimize.jl")
 
 # Fitness
-include("fitness/fitness_types.jl")
+# include("fitness/fitness_types.jl") FIXME merge it with fitness.jl
 include("fitness/pareto_dominance.jl")
 include("fitness/epsilon_pareto_dominance.jl")
 include("fitness/epsilon_box_dominance.jl")
--- a/src/adaptive_differential_evolution.jl
+++ b/src/adaptive_differential_evolution.jl
@@ -41,22 +41,21 @@ function adjust!( params::AdaptiveDiffEvoParameters, index, is_improved::Bool )
     end
 end
 
-function adaptive_diffevo(name::ASCIIString,
+function adaptive_diffevo(problem::OptimizationProblem, name::ASCIIString,
                  crossover::DiffEvoCrossoverOperator,
                  select::IndividualsSelector = SimpleSelector(),
                  options = @compat Dict{Symbol,Any}())
   opts = Parameters(options, ADE_DefaultOptions)
-  ss = opts[:SearchSpace]
   pop = opts[:Population]
   DiffEvoOpt(name, pop, AdaptiveDiffEvoParameters(opts, popsize(pop)), select,
-        NoMutation(), crossover, RandomBound(ss))
+        NoMutation(), crossover, RandomBound(search_space(problem)))
 end
 
-adaptive_de_rand_1_bin(options = @compat(Dict{Symbol,Any}()),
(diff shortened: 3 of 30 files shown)


robertfeldt:
Looks good and improves fitness overall. Some problems with generating set search though:

````
ERROR: `RandomBound{S<:SearchSpace}` has no method matching RandomBound{S<:SearchSpace}(::Bool)
 in GeneratingSetSearcher at /Users/feldt/tmp/BlackBoxOptim.jl/src/generating_set_search.jl:66
 in direct_search_probabilistic_descent at /Users/feldt/tmp/BlackBoxOptim.jl/src/direct_search_with_probabilistic_descent.jl:52
 in setup_bboptimize at /Users/feldt/tmp/BlackBoxOptim.jl/src/bboptimize.jl:243
 in bboptimize at /Users/feldt/tmp/BlackBoxOptim.jl/src/bboptimize.jl:175
 in fitness_for_opt at /Users/feldt/tmp/BlackBoxOptim.jl/examples/benchmarking/compare_optimizers.jl:192
 in compare_optimizers_to_benchmarks at /Users/feldt/tmp/BlackBoxOptim.jl/examples/benchmarking/compare_optimizers.jl:349
 in main at /Users/feldt/tmp/BlackBoxOptim.jl/examples/benchmarking/compare_optimizers.jl:116
 in include at /usr/local/Cellar/julia/0.3.10/lib/julia/sys.dylib
 in process_options at /usr/local/Cellar/julia/0.3.10/lib/julia/sys.dylib
 in _start at /usr/local/Cellar/julia/0.3.10/lib/julia/sys.dylib
while loading /Users/feldt/tmp/BlackBoxOptim.jl/examples/benchmarking/compare_optimizers.jl, in expression starting on line 52
```

alyst:
How can I reproduce? Is it enough just to run ```compare_optimizers.jl```?

robertfeldt:
Or probably any run with gss. The runcompare script should do it but do
numreps 1.

fredag 10 juli 2015 skrev Alexey Stukalov <notifications@github.com>:

> How can I reproduce? Is it enough just to run compare_optimizers.jl?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/robertfeldt/BlackBoxOptim.jl/pull/18#issuecomment-120501122>
> .
>


-- 
Best regards,

/Robert Feldt
--
Tech. Dr. (PhD), Professor of Software Engineering
Blekinge Institute of Technology, Software Engineering Research Lab, and
Chalmers, Software Engineering Dept
Explanea.com - Igniting your Software innovation
robert.feldt (a) bth.se    or    robert.feldt (a) chalmers.se    or
 robert.feldt (a) gmail.com
Mobile phone: +46 (0) 733 580 580
http://www.robertfeldt.net <http://www.cse.chalmers.se/~feldt>

robertfeldt:
We should add short runs of all optimizers to test suite. I'm on run right now but can fix later...

alyst:
Should be fixed now (I've tested on v0.4, but v0.3 should work too, I hope). I've replaced my branch, so probably you would have to delete your local version again. Sorry for the inconvenience.

multidis:
Fitness storage is a helpful starting point for some cases when objective function is expensive and could be replaced by e.g. Gaussian Process regression within some acceptable confidence intervals based on already stored points. Thanks @alyst

robertfeldt:
Looking good and running the benchmark runs now. Will then merge. Have a few comments/concerns maybe you can comment on:

* Not convinced that VectorFitness is the way to go. For example, in a pareto fitness scheme one cannot first aggregate and then compare, one must compare the vectors directly. Do you think this should be a separate fitness scheme/type or how do you envisage this? I think the representation and the (scheme) that interprets the representation should be different. Actually it can be very complex with very many different levels of evaluation.

* I'm also a bit hesitant about last_fitness since this makes parallel evaluation very hard. Maybe we should change the design so that one can evaluate a candidate solution and get back a fitness object directly rather than have a two-step procedure. Or have I misunderstood what you are doing here?

* I will have to check but I think the xNES and separable_NES do not yet embed in the searchspace so they might return invalid solutions. Possible issue to fix for the future, will note it in separate issue.

* UnboundedProblem and BoundedProblem more implies whether they have bounding of the search space for me. Can we find better terms for known optval?

alyst:
> Not convinced that VectorFitness is the way to go.

Maybe it's my misunderstanding of ```aggregate()``` and ```RatioFitnessScheme``` purpose. I wasn't sure how it should be used, so I tried to keep it as intact as possible. But the proposed design of ```FitnessScheme```, I believe, allows for proper comparison of vector fitnesses using the domination ordering throughout the optimization process, the details could be sorted out when we would have some example problems.

What I'm not sure about is whether this PR fits the ideas drafted in ```fitness/fitness_types.jl```. I have no experience of such extensions of fitness concept, so it might require few more design iteration(s).

As a side note wrt vector fitness I just thought that the we can use ```NTuple{T,N}``` (introduced recently in v0.4) as the fitness type -- it would act as a vector with a fixed length known at compile time.

> I'm also a bit hesitant about last_fitness since this makes parallel evaluation very hard.

Thanks for noticing. That was a fragment of the older code. Now I fixed it.

> ```UnboundedProblem``` and ```BoundedProblem``` more implies whether they have bounding of the search space for me.

I agree. Are e.g. ```UnknownOptimumProblem``` and ```KnownOptimumProblem``` better?

robertfeldt:
Ok, great. We can sort out the fitness thing later; the thing sketched out in fitness_types is base don state-of-the-art research so we can sort it out later. :)

The results from the benchmark comparison looks really good:

```
Num problems (any p-value) = 159
  Num where new implementation is better = 123
  Num where new implementation is worse = 36
Num problems (with p-values < 0.05) = 76
  Num where new implementation is better = 75
  Num where new implementation is worse = 1
Num problems (with p-values < 0.01) = 74
  Num where new implementation is better = 74
  Num where new implementation is worse = 0
```
So I'll go ahead and merge then. Thanks a lot, getting there when it comes to clean code etc.

multidis:
Factor of 2 spedup observed indeed with the fixed `:MaxSteps` setting after this merge (tested for `:adaptive_de_rand_1_bin_radiuslimited`).

robertfeldt:
Yes, and the new benchmark result file shows adaptive_de_rand_1_bin_radiuslimited is now on top on mean rank. It's the most robust choice of the ones tested so far. I plan to add some more optimizers soon.

robertfeldt:
https://github.com/robertfeldt/BlackBoxOptim.jl/blob/master/examples/benchmarking/latest_toplist.csv

@multidis

multidis:
I do see now some tests failing after the latest commits but so far those seem to be due to a somewhat less precise optimum coordinates (some simpler cases e.g. Rosenbrock 2D), global optima are still being found. Will have to decrease the tol for those tests. Local search refinement should help.

robertfeldt:
I'm very interested if you can provide example code for those so we can include in future benchmarks. We want to be sure not to regress in fitness performance.

multidis:
context("BlackBoxOptim bound-constrained global, Rosenbrock 2D") do
    optres = BlackBoxOptim.bboptimize(
      rosenbrock, search_range = [(-10.0, 10.0), (-10.0, 10.0)],
      method = :adaptive_de_rand_1_bin_radiuslimited,
      parameters = {:MaxSteps => 10_000, :ShowTrace => false}
      )
    println(optres[1])
    @fact norm(optres[1] - [1.0, 1.0]) < xnormtol => true
  end

function rosenbrock(x::Array)
  # global min at (1,1), f(1,1)=0
  return (1.0 - x[1])^2 + 100.0*(x[2] - x[1]^2)^2
end
function rastrigin3d(x::Array)
  # global min at xi=0 for all i
  return 30.0 + sum(x.^2 - 10.0*cos(2pi*x))
end

My previous tests had `xnormtol = 1e-5`

New failing run
[0.9999793779487288,0.9999593448316044]

robertfeldt:
Ah ok, we test with Rosenbrock quite extensively but not in 2D. Hard to understand why it cannot solve that better in 10k steps, should be easy compared to what we throw at it. Anyway, testing bbopt's is a probabilistic game so har to analyse properly.

multidis:
Some performance regression on more complex test cases (private code) but the Rosenbrock example is fairly representative.

multidis:
Sorry take the last 4D statement back - the tol for checks was much lower so it still passes.

multidis:
Now I am hesitating about the last statement (about fair price): some more serious failures in tests where results depend on post-processing a series of optimizations. Qualitatively incorrect results at times. Have to revert back to the May version of the package.

alyst:
@multidis There were some serious errors with the Archive class (see #20). So if your code took the results from the archive, that could be the reason for regressions. The archive should be fixed in the latest master.

multidis:
@alyst unfortunately the failures are observed with the latest master, just updated and checked again.

Trying currently to identify now the latest commit that passes my tests.

robertfeldt:
Sad to hear that. We have benchmarked extensively in each step but I have only included 5-dimensional problems and up. Hmm, will go back and add a few 2-dim ones to see if I see similar problems there.

alyst:
@multidis Thanks for looking into this. The logic of ADE should have been untouched by the recent changes, so it must be some bug. Would be nice to bisect the responsible commit.

multidis:
The latest group of commits from July 12 that were merged as **fix archive** are the ones that end up producing a number of failed tests. July 11 commit 61e853e38a1549612e0debc7efff658e5ed44b85 passes all my tests.

Some **background**: All test cases are up to 4-dim. Testes with `:adaptive_de_rand_1_bin_radiuslimited` only, `10_000` steps in each optimization run. The objective functions generally have a large number of local minima fairly close to the global ones.

Picking local minima has profound effects in generating unphysical parameters so it is more than just a matter of some accuracy degradation. From a set of runs I checked with multiple commits, it appears (so far without exceptions) that the breaking commits degrade Rosenbrock 2D accuracy (as in my example above) while the passing ones return `[1, 1]` with at lest `1e-8` precision.

multidis:
I could try to sort out which specific commit from the last merge is the breaking one if that helps; so far tried a couple of those (both failing) but a bit confused with the order of those commits (is there a reason github lists older ones by date after more recent ones in the commit tree?) @alyst

alyst:
> Is there a reason github lists older ones by date after more recent ones in the commit tree?
That's because I was patching the older commits -- github sorting reflects the order at which the changes should be applied.

>I could try to sort out which specific commit from the last merge is the breaking one if that helps; so far tried a couple of those (both failing)
I have tried to make the commits atomic, but some of them might fail, sorry for that.
What Julia version are you using?

multidis:
Oh, I meant fail the tests, not fail to run, sorry if confusing.
Running Julia 0.3.10

alyst:
@multidis If you could dissect which commit within #20 is working for you it would help a lot, because I mostly use v0.4 and there were some version-specific changes in that PR.

multidis:
Did that: the very first commit from that group, 228285b7599deaa2e519aee9f0f3279c2928bb01 "fix delta_fitness..." fails the tests.

I'd be really interested to know if that is a version specific issue, which would imply things quietly breaking as I upgrade to 0.4 when stable...

alyst:
@multidis Thanks! Do your failing tests stop before the MaxTime/MaxSteps are reached? This commit affects the termination condition, before the patch the fitness delta was not really changing during the optimization.

multidis:
This is a good question @alyst : not sure, I have not explored the stopping condition change. My function call is

    optglob = BlackBoxOptim.bboptimize(
      fobj, search_range = parconstr,
      method = :adaptive_de_rand_1_bin_radiuslimited,
      parameters = {:MaxSteps => adaptive_de_maxsteps,
                    :PopulationSize => ga_population_size,
                    :ShowTrace => false}
      )

My intent is to run 10_000 steps, that was an empirically established parameter from the times when there were no stopping condition options in BlackBoxOptim.
Does the new merge involve a termination condition that is fired with such a function call?

alyst:
@multidis The merge implicitly affects the termination condition
```delta_fitness(evaluator.archive) < parameters[:MinDeltaFitnessTolerance]```.
It was not really checked before (always checked the very first fitness improvement). By default ```:MinDeltaFitnessTolerance => 1e-11```, you can try reducing it to zero.

**Comment by multidis** [comment]
Now I see: the Rosenbrock example verbose run:

Optimization stopped after 5018 steps and 0.14467716217041016 seconds
Termination reason: Delta fitness below tolerance

But not after 10_000 steps...

**Comment by multidis** [comment]
All the tests pass when setting `:MinDeltaFitnessTolerance => 0.0`. Still factor of 2 speedup, per 10_000 (not 20_000) function evaluations. Thank you @alyst for the stopping condition pointer.

Regarding the stopping tol (which is good to have as an option anyway): is the fitness delta based on 2 successive evaluations or some longer history period?

**Comment by alyst** [comment]
@multidis Great that it's resolved. Thanks for you help.
As it is defined now, ```delta_fitness()``` is the difference between the current best fitness and the former best fitness (not the best fitness of the previous iteration, but from some iteration in the past when it was different from the current fitness).

**Comment by multidis** [comment]
I see, no further improvement condition, but within how many steps/function evaluations?

Say fitness evaluated at func. call number 4000 was the former best fitness and up to 5018 steps (Rosenbrock output above) any new best fitness candidate was within tol to that one. How is the termination step determined in a sense that the run proceeded long enough, 1018 steps, to see no improvement?

**Comment by alyst** [comment]
In that example let's also assume that at iteration 3000 the best fitness was 1.3E-10, then no improvements for 1000 iterations (the best fitness is exactly constant) until it became 1E-11 at 4000-th iteration. So from 4000-th to 5018-th iteration the delta fitness would be 2E-11.
If best fitness at 5018 step is e.g. 1E-12, then delta would become 9E-12 and the stopping condition would be satisfied.
So this stopping condition is unstable in a sense that any small change early in the optimization process could terminate it.

**Comment by robertfeldt** [comment]
Ok, sounds like you cleared this up then. Are we ok with the current termination logic or do you propose any changes? Thanks to both of you for clearing this up while I was asleep... :)

**Comment by robertfeldt** [comment]
I went back and run low-dimensional (2D) benchmarks on some problems at an end-of-May commit and then rerun with latest master:

```
Num problems (any p-value) = 35
  Num where new implementation is better = 14
  Num where new implementation is worse = 21
Num problems (with p-values < 0.05) = 6
  Num where new implementation is better = 0
  Num where new implementation is worse = 6
Num problems (with p-values < 0.01) = 6
  Num where new implementation is better = 0
  Num where new implementation is worse = 6
6x9 DataFrame
| Row | Problem     | NumDims | Method                        | OldMedian  |
|-----|-------------|---------|-------------------------------|------------|
| 1   | "Griewank"  | 2       | "probabilistic_descent"       | 0.00739604 |
| 2   | "Rastrigin" | 2       | "probabilistic_descent"       | 4.59941e-9 |
| 3   | "Griewank"  | 2       | "generating_set_search"       | 0.00739604 |
| 4   | "Rastrigin" | 2       | "generating_set_search"       | 4.39651e-9 |
| 5   | "Griewank"  | 2       | "de_rand_1_bin_radiuslimited" | 5.8569e-9  |
| 6   | "Ackley"    | 2       | "probabilistic_descent"       | 7.17668e-9 |

| Row | Order | NewMedian  | Pvalue      | Level | StatSignDiff |
|-----|-------|------------|-------------|-------|--------------|
| 1   | "<"   | 0.245353   | 7.73255e-11 | 0.01  | "Yes"        |
| 2   | "<"   | 7.95966    | 1.75645e-10 | 0.01  | "Yes"        |
| 3   | "<"   | 0.0986221  | 6.32034e-9  | 0.01  | "Yes"        |
| 4   | "<"   | 1.98992    | 1.84999e-8  | 0.01  | "Yes"        |
| 5   | "<"   | 0.00739604 | 0.00159665  | 0.01  | "Yes"        |
| 6   | "<"   | 2.57993    | 0.00485602  | 0.01  | "Yes"        |
```

So it seems there has been some effects that are unlikely to be because of chance. Only DE affected is de_rand_1_bin_radiuslimited on Griewank though. Hard to find a reason that probabilistic descent or generating set search would be affected since changes have focused on DE. Can be because of changes to termination logic. @multidis if you have specific problems where you saw problems we can add them to BBO. Rosenbrock we already have but we are missing Tsallis. Any other you test with that you can share?

**Comment by alyst** [comment]
```GeneratingSetSearcher``` also uses the archive, so I would say ```delta_fitness()``` termination condition is the primary suspect.
As for the termination condition, I would say that something like running average of fitness improvement should be more stable.

**Comment by robertfeldt** [comment]
Yes, likely. When I rerun with

```
:MinDeltaFitnessTolerance => 1e-50
```
or 0.0 I get infinite runs since num fevals is not increasing on Griewank 2D though:

```
Griewank, n = 2, optimizer = de_rand_1_bin_radiuslimited, run 3 of 30 (26.0% of total runs)
Starting optimization with optimizer DE/rand/1/bin/radiuslimited
0.50 secs, 7785 evals, 32126 steps, improv/step: 0.050 (last = 0.0499), 0.007396040
1.00 secs, 7785 evals, 66725 steps, 0.007396040
1.50 secs, 7785 evals, 102124 steps, 0.007396040
2.00 secs, 7785 evals, 136814 steps, 0.007396040
2.50 secs, 7785 evals, 171757 steps, 0.007396040
3.00 secs, 7785 evals, 205780 steps, 0.007396040
3.52 secs, 7785 evals, 241666 steps, 0.007396040
4.02 secs, 7785 evals, 277561 steps, 0.007396040
4.53 secs, 7785 evals, 313456 steps, 0.007396040
5.05 secs, 7785 evals, 349350 steps, 0.007396040
5.55 secs, 7785 evals, 385245 steps, 0.007396040
6.09 secs, 7785 evals, 421139 steps, 0.007396040
6.60 secs, 7785 evals, 457034 steps, 0.007396040
7.12 secs, 7785 evals, 492929 steps, 0.007396040
7.64 secs, 7785 evals, 528823 steps, 0.007396040
8.14 secs, 7785 evals, 564755 steps, 0.007396040
8.66 secs, 7785 evals, 600613 steps, 0.007396040
9.17 secs, 7785 evals, 636507 steps, 0.007396040
9.69 secs, 7785 evals, 672402 steps, 0.007396040
```

Strange, will debug.

**Comment by alyst** [comment]
The fitness is reset (and then recalculated) if the trial is different from target. It could be that either cr==0.0 or the population is degenerated (all the individuals are the same).

**Comment by robertfeldt** [comment]
Yes, seems likely it was the latter. I think this can happen for many reasons so better to add a termination criterion for it. I did so in main loop and a param to set max num steps without a func eval (currently 100).

**Comment by robertfeldt** [comment]
With this change there is no detected performance regressions for low-dim problems between late-May and current master:

```
Num problems (any p-value) = 35
  Num where new implementation is better = 12
  Num where new implementation is worse = 23
Num problems (with p-values < 0.05) = 1
  Num where new implementation is better = 0
  Num where new implementation is worse = 1
1x9 DataFrame
| Row | Problem    | NumDims | Method                        | OldMedian |
|-----|------------|---------|-------------------------------|-----------|
| 1   | "Griewank" | 2       | "de_rand_1_bin_radiuslimited" | 5.8569e-9 |

| Row | Order | NewMedian  | Pvalue    | Level | StatSignDiff |
|-----|-------|------------|-----------|-------|--------------|
| 1   | "<"   | 0.00739604 | 0.0359805 | 0.01  | "No"         |
Num problems (with p-values < 0.01) = 0
  Num where new implementation is better = 0
  Num where new implementation is worse = 0
```

Note that the 0.05 alpha level is not very reliable here when we do 7*5=35 comparisons.

multidis:
> if you have specific problems where you saw problems we can add them to BBO. Rosenbrock we already have but we are missing Tsallis. Any other you test with that you can share?

Among the closed-form analytical objective functions I am testing Rosenbrock 2D, Rastrigin 3D, and this one (probably not among canonical test problems for optimization but useful in a domain-specific way) in 4D with one global and 15 degenerate local minima:

    function etsallis(x::Array)
      return sumabs2(x.^2 - 8.0) + 5.0*sum(x) + 57.3276
    end

From C. Tsallis, Physica A 233 (1996) 395-406 (section 3 eq.(22)).

Another possibly useful set is table 1 of Xiang et al., The R Journal Vol. 5/1, June 2013, p.13-28. That paper actually made me abandon differential evolution in favor of simulated annealing in 2013, but BlackBoxOptim.jl switched things back in 2014 (also GenSA C++ code is not as clean as I'd like it to be IMHO).

PS: The latest commit 079f886764649ea9521f844e50ce43be7bbc1d9a passes all non-closed-form objective tests as well.

PR #20, opened 2015-07-12
Fix Archive

alyst:
After the last update ```TopListArchive``` was badly broken: it stored the references to candidate parameters, but since they were returned to the pool and reused, the archive contents were totally wrong. Also the archive was not respecting the fitness scheme. This should be fixed now.

I've also renamed the fitness scheme types, e.g. ```ScalarFitness``` to ```ScalarFitnessScheme``` otherwise someone might confuse it with the actual fitness values. Now there's also ```MinimizingFitnessScheme``` and ```MaximizingFitnessScheme``` constants for readability.


alyst:
added DE patch here because it looks trivial

PR #21, opened 2015-07-17
Parameters enhancements

alyst:
Some cleaning up for the Dictionary-based parameters.
 * ```Parameters``` class renamed into ```DictChain``` and parameterized by key and value types. ```Parameters``` now is an alias to ```DictChain{Symbol,Any}``` (so all method parameters must be symbols). This should help identifying the issues with incorrect key types at compile time.
 * ```mergeparam()``` renamed into ```chain()```, but there's also ```Base.merge()``` methods
 * ```getindex()``` throws ```KeyError``` if the key is missing. Since we should have the default value for every method parameter in ```DictChain```, any request of a missing key is actually a bug.
 * ```convert(Dict, DictChain)``` method. I had some obscure problems with Parameters type serialization (for parallelization), using dict solved it.

### DIFF
--- a/src/BlackBoxOptim.jl
+++ b/src/BlackBoxOptim.jl
@@ -13,7 +13,7 @@ export  Optimizer, AskTellOptimizer, SteppingOptimizer, PopulationOptimizer,
         XNESOpt, xnes,
 
         # Parameters
-        Parameters, mergeparam,
+        DictChain, Parameters, mergeparam,
 
         # Fitness
         FitnessScheme,
--- a/src/BlackBoxOptim.jl
+++ b/src/BlackBoxOptim.jl
@@ -13,7 +13,8 @@ export  Optimizer, AskTellOptimizer, SteppingOptimizer, PopulationOptimizer,
         XNESOpt, xnes,
 
         # Parameters
-        DictChain, Parameters, mergeparam,
+        DictChain, Parameters,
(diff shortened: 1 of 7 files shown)


alyst:
@robertfeldt Thanks! I have 3 PRs more: OptController class (#22), enabling frequency adaptation for mixtures of genetic operators and some sort of parallelization for population optimizers.

multidis:
Breaking changes with the latest master-commit

ERROR: `chain` has no method matching chain(::Dict{Symbol,Any}, ::Dict{Any,Any})
 in setup_bboptimize at .../.julia/v0.3/BlackBoxOptim/src/bboptimize.jl:188

Do we need to change the call to bboptimize?

alyst:
@multidis I guess the call looks like ```bboptimize(..., {:MaxSteps => 5, ...})```.
The problem is that ```{...}``` generates ```Dict{Any,Any}``` and is deprecated in v0.4. You have to use ```Compat``` package and change ```{...}``` into ```@compat(Dict{Symbol,Any}(...))```

robertfeldt:
Hmm, lets fix this so we can use also the old format. I missed it in my testing before merge.

alyst:
@robertfeldt I've just pushed the changes allowing Dict{Any,Any}, but with the deprecation warning (displayed once).

PR #22, opened 2015-07-17
OptController class

alyst:
There's some cosmetic fixes here and there, but the major change is the introduction of ```OptController``` class.
Basically, I have moved the ```run_optimizer!()``` method into a separate class and extracted some functions, like ```check_stop_conditions()``` or ```trace_progress()```. Besides improving code structure, this class could serve as low-level interface for running optimizations with custom Optimizer and Evaluator instances.

### DIFF
--- a/examples/benchmarking/compare_optimizers.jl
+++ b/examples/benchmarking/compare_optimizers.jl
@@ -3,6 +3,7 @@ using DataFrames
 using BlackBoxOptim
 using ArgParse
 using CPUTime
+using Compat
 
 global logfilehandle = nothing
 function log(color::Symbol, str)
@@ -129,7 +130,7 @@ function main(args)
   close(logfilehandle)
 end
 
-ProblemSets = {
+ProblemSets = @compat Dict{ASCIIString,Any}(
   "easy" => [
     # ProblemName, NumDims, PopSize, MaxFevals
     ("Sphere",        5, 20, 5e3),
@@ -172,21 +173,21 @@ ProblemSets = {
     ("Rosenbrock",    2, 25, 1e4),
     ("Rastrigin",     2, 25, 1e4),
     ("Ackley",        2, 25, 1e4),
-    ("Griewank",      2, 25, 1e4),  
+    ("Griewank",      2, 25, 1e4),
   ],
 
   "test" => [
     ("Rosenbrock",   30, 50, 2e5),
   ]
-}
+)
 ProblemSets["all"] = vcat(ProblemSets["easy"], ProblemSets["harder"])
 
-OptimizerSets = {
+OptimizerSets = @compat Dict{ASCIIString,Any}(
   "de" => [:de_rand_1_bin, :de_rand_1_bin_radiuslimited, :adaptive_de_rand_1_bin, :adaptive_de_rand_1_bin_radiuslimited],
   "stable_non_de" => [:probabilistic_descent, :generating_set_search, :random_search],
   "nes" => [:xnes, :separable_nes],
   "test" => [:de_rand_1_bin],
-}
+)
 OptimizerSets["all"] = collect(keys(BlackBoxOptim.ValidMethods))
 OptimizerSets["stable"] = vcat(OptimizerSets["de"], OptimizerSets["stable_non_de"])
 
--- a/examples/benchmarking/compare_optimizers.jl
+++ b/examples/benchmarking/compare_optimizers.jl
@@ -200,17 +200,17 @@ increase_runs_per_problem(problemname, numdims) = begin
   RunsPerProblem[k] = get(RunsPerProblem, k, 0) + 1
 end
 
-function fitness_for_opt(family::FunctionBasedProblemFamily, numDimensions, populationSize, numFuncEvals,
-    method, showtrace = true)
+function fitness_for_opt(family::FunctionBasedProblemFamily, numDimensions::Int, populationSize::Int, numFuncEvals::Int,
+    method::Symbol, showtrace::Bool = true)
 
   problem = BlackBoxOptim.fixed_dim_problem(family, numDimensions)
 
-  best, fitness = bboptimize(problem; method = method, parameters = {
+  best, fitness = bboptimize(problem; method = method, parameters = @compat(Dict{Symbol,Any}(
     :NumDimensions => numDimensions,
     :PopulationSize => populationSize,
     :ShowTrace => showtrace,
     :MaxFuncEvals => numFuncEvals
-    })
+    )))
 
   fitness
 end
@@ -246,7 +246,7 @@ function multitest_opt(problemDescriptions, method; NumRepetitions = 3)
 
       start_time = time()
       CPUtic()
-      ftn = fitness_for_opt(prob, numdims, popsize, numfevals, method)
+      ftn = fitness_for_opt(prob, numdims, popsize, ceil(Int, numfevals), method)
       df[:ElapsedTime] = CPUtoc()
       df[:StartTime] = strftime("%Y-%m-%d %H:%M:%S", start_time)
       df[:Fitness] = ftn
@@ -366,7 +366,7 @@ function compare_optimizers_to_benchmarks(benchmarkfile, pset, optimizers, nreps
                 for r in 1:nreps
                     runnum += 1
                     log("\n$(probname), n = $(numdims), optimizer = $(string(optmethod)), run $(r) of $(nreps) ($(round(100.0 * runnum / totalruns, 2))% of total runs)\n")
-                    ftn = fitness_for_opt(prob, numdims, popsize, numfevals, optmethod)
+                    ftn = fitness_for_opt(prob, numdims, popsize, ceil(Int, numfevals), optmethod)
                     push!(newfs, ftn)
                 end
                 pval = pvalue(MannWhitneyUTest(benchfitnesses, newfs))
--- a/src/BlackBoxOptim.jl
+++ b/src/BlackBoxOptim.jl
@@ -98,6 +98,7 @@ include(joinpath("problems", "all_problems.jl"))
 include(joinpath("problems", "problem_family.jl"))
 
 include("evaluator.jl")
+include("opt_controller.jl")
 
 function setup(o::Optimizer, evaluator::Evaluator)
   # Do nothing, override if you need to setup prior to the optimization loop
--- a/src/BlackBoxOptim.jl
+++ b/src/BlackBoxOptim.jl
@@ -13,7 +13,7 @@ export  Optimizer, AskTellOptimizer, SteppingOptimizer, PopulationOptimizer,
         XNESOpt, xnes,
 
         # Parameters
-        DictChain, Parameters,
+        DictChain, Parameters, ParamsDictChain, ParamsDict,
         chain,
 
         # Fitness
--- a/src/BlackBoxOptim.jl
+++ b/src/BlackBoxOptim.jl
@@ -100,11 +100,11 @@ include(joinpath("problems", "problem_family.jl"))
 include("evaluator.jl")
 include("opt_controller.jl")
 
-function setup(o::Optimizer, evaluator::Evaluator)
+function setup!(o::Optimizer, evaluator::Evaluator)
   # Do nothing, override if you need to setup prior to the optimization loop
 end
 
-function finalize(o::Optimizer, evaluator::Evaluator)
+function finalize!(o::Optimizer, evaluator::Evaluator)
   # Do nothing, override if you need to finalize something after the optimization loop
 end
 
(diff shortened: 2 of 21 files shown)


robertfeldt:
Looks nice but I want to benchmark. Seems there is one more update needed in bboptimize.jl:

```
feldt:~/tmp/BlackBoxOptim.jl/examples/benchmarking$ julia --color=yes -L ../../src/BlackBoxOptim.jl compare_optimizers.jl compare --benchmarkfile benchmark_runs_150712.csv -n 2

Sphere, n = 5, optimizer = de_rand_1_bin, run 1 of 2 (0.31% of total runs)
ERROR: type: setup_bboptimize: in typeassert, expected Float64, got Bool
 in bboptimize at /Users/feldt/tmp/BlackBoxOptim.jl/src/bboptimize.jl:174
 in fitness_for_opt at /Users/feldt/tmp/BlackBoxOptim.jl/examples/benchmarking/compare_optimizers.jl:207
 in fitness_for_opt at /Users/feldt/tmp/BlackBoxOptim.jl/examples/benchmarking/compare_optimizers.jl:205
 in compare_optimizers_to_benchmarks at /Users/feldt/tmp/BlackBoxOptim.jl/examples/benchmarking/compare_optimizers.jl:368
 in main at /Users/feldt/tmp/BlackBoxOptim.jl/examples/benchmarking/compare_optimizers.jl:123
 in include at /usr/local/Cellar/julia/0.3.10/lib/julia/sys.dylib
 in process_options at /usr/local/Cellar/julia/0.3.10/lib/julia/sys.dylib
 in _start at /usr/local/Cellar/julia/0.3.10/lib/julia/sys.dylib
while loading /Users/feldt/tmp/BlackBoxOptim.jl/examples/benchmarking/compare_optimizers.jl, in expression starting on line 52
```

alyst:
Sorry, I didn't test ```compare_optimizers.jl``` against #21. Should be fixed now.
Also in light of @multidis comments I've added more did another iteration on parameters, so that incorrect type is detected as early as possible. But now ```Parameters``` itself is an abstract type (```Associative{Symbol,Any}```).

robertfeldt:
Seems it is not using the MaxFuncEvals param but rather running until fitness within tolerance 1e-8. Started this run with MaxFuncEvals => 3e5 but got:

```
...
Optimization stopped after 25298958 steps and 271.6293931007385 seconds
Termination reason: Fitness (6.692224147638105e-9) within tolerance (1.0e-8) of optimum
Steps per second = 93137.77758439213
Function evals per second = 93138.40711861906
Improvements/step = Inf
Total function evaluations = 25299129
```

robertfeldt:
It seems to me the problem is in setup_bboptimize sinec max_time has value 0.0 by default, it is entered into Parmaters and then sets MaxFuncEvals to 0 on line 208 which makes the check_stop_condition(ctrl::OptController) not check max_fevals. Please fix this since I'm unsure with your proposed changes what is the right order of things here.

robertfeldt:
Yes, verified that this fixes the problem with MaxFuncEvals:

```
diff --git a/src/bboptimize.jl b/src/bboptimize.jl
index 7f32407..f70826f 100644
--- a/src/bboptimize.jl
+++ b/src/bboptimize.jl
@@ -204,9 +204,11 @@ function setup_bboptimize(functionOrProblem; max_time::Number = 0.0,
     if !isa(params[:MaxTime], Number) || params[:MaxTime] < 0.0
       throw(ArgumentError("max_time parameter must be a non-negative number"))
     else
-      params[:MaxTime] = convert(Float64, params[:MaxTime])
-      params[:MaxFuncEvals] = 0
-      params[:MaxSteps] = 0
+      if params[:MaxTime] > 0.0
+        params[:MaxTime] = convert(Float64, params[:MaxTime])
+        params[:MaxFuncEvals] = 0
+        params[:MaxSteps] = 0
+      end
     end
   end
```

Still, maybe there is a better way to clean up this logic given your changes?

alyst:
@robertfeldt Thanks for looking into it. Yes, that was the problem. I think the parameter logic is a little bit fragile because we have both keyword arguments and parameters dictionary plus :MaxTime, :MaxFuncEvals priorities on top.
I have pushed the fix, also removed the keyword arguments from ```setup_bboptimize()```, now ```bboptimize()``` is the only function with keyword arguments.

alyst:
Removing the keywords arguments from ```bboptimize()``` might help simplifying it. I think these arguments are more tuned for the problems of the benchmarking suite, and for real problems the user will have to touch the parameters dictionary anyway. But we also need to somehow expose to use user the options required by ```bboptimize()``` and each specific method.

robertfeldt:
But I want the top-level function to make the most common use case as easy as possible. Maybe we can allow keyword arguments but they should have the same names as they have in the param dict? And only make exception for the different ways to set up the search space?

alyst:
Having the same names for parameters would definitely help. IIRC Julia allows arbitrary keyword arguments via ```bboptimize(x; kwargs...)``` syntax, where ```kwargs``` would be a dictionary inside the function. The function itself could be just a wrapper that calls dictionary-based ```bboptimize()```.

robertfeldt:
Unfortunately I'm seeing a bit too many worse results in benchmarking so need to investigate more if this is a real effect. Seems mostly generating_set_search and probabilistic descent are affected. Can you see any reason why? Some parameter changed or is not being picked up?

```
Num problems (any p-value) = 160
  Num where new implementation is better = 57
  Num where new implementation is worse = 103
Num problems (with p-values < 0.05) = 33
  Num where new implementation is better = 3
  Num where new implementation is worse = 30
Num problems (with p-values < 0.01) = 27
  Num where new implementation is better = 1
  Num where new implementation is worse = 26
```

alyst:
That's bad. Do you know what's the termination reason? Does it do fewer function evals per second?

alyst:
I am still not sure about the reason, but one hypothesis is that parameters lookup become slower and GSS uses parameters inside ```step!()``` a lot. So I've replaced parameters lookup by the fields in GSS type, which is a right thing to do anyway.

robertfeldt:
Hmm, not sure, I did not save the termination reason. Will check. Don't think it can be this since I benchmark on a fixed set of evals and only terminate if within 1e-8 of known optimum.

robertfeldt:
Seems it can be the old MinDeltaFitness bug again. I'm sure I had fixed that in compare_optimizers.jl. Not sure how current commits are missing it. Anyway, will do some more testing to see if this is really it. If it is I'll merge and then set default value of MinDeltaFitness to 0.0 or something really low.

alyst:
Thanks for helping to sort out the issues with this PR. The last two pull requests were not adding new features, but hopefully they will help structuring the code better.

robertfeldt:
@alyst How do you view the OptController in relation to an OptimizationState similar to Optim.jl. My thinking is that we might want to merge them into one or do you see the controller as only handling the main opt loop? I would like to add OptimizationState and/or OptimizationResult type(s) but maybe they can also be merged? There are a couple of things I want to achieve in one go:

1. Avoid sending a tuple back since we might want to send back a lot of different result- and state-related info to caller.

2. Allowing caller to stop the optimization early. Possible design: A Parameter "UserTerminationFunction" (or some better name ;) ) can be set to a function which will then be called at end of each opt step, passing in an OptState so that the user can manually decide to stop the opt run premateurely.

3. Allowing the caller to continue an optimization run after it has finished.

Would be great to fix these things in one go and at the same time change the top-level interface to have a more unified way of specifying parameters. Ideas and comments?

alyst:
@robertfeldt Definitely API compatibility with Optim.jl is very important. I'm not sure whether current Optim API is finalized, but it doesn't seem a big deal to sync if it changes. There are some BlackBox or even method-specific features missing in ```OptimizationResult``` and some (i.e. gradient convergence) that do not apply to BlackBox case, so while it would be nice to have ```OptimizationResult```-derived type, maybe the compatibility should be on the function level, rather than type level, i.e. Optim should declare ```num_iterations(result)``` or ```x_converged(result)``` functions and BlackBoxOptim should support/implement corresponding methods for its result type.

I think it's nice to have ```OptController``` to handle the main loop and serve as "the optimizer object" to the end user. Potentially it allows to reuse the same optimizer object for different input data (IIUC correctly that's something that is missing in Optim: if one wants to repeatedly apply the same Optim solver, there would be some initialization overhead of closures construction etc) and resuming the optimization.

The question of ```OptimizationState``` is related a bit to the design of ```Evaluator``` and ```Archive``` classes. If we are saving traces from each iteration (which is critical for research, I suppose), would you do this in ```Evaluator```.```Archive``` or you would rather want ```Archive``` to be the storage of the current best result, and traces should be saved in another structure?
As with ```OptimizationResult```, it would make sense to extend ```OptimizationState```/```OptimizationTrace``` to store some BBO-specific information, e.g. the genetic operation applied and remove things like gradient norm. The ```OptimizationState```-compatible structure could also be an input to ```UserTerminationFunction```` (or ```OptStepCallback```?). This function (I would allow it to be any callable object, not just ```Function```) could then be used for tracing as well.

robertfeldt:
@alyst @multidis Please check the proposed top-level use cases shown here:

https://github.com/robertfeldt/BlackBoxOptim.jl/blob/master/design/design_by_testcode/test_bboptimize_toplevel_interface.jl

I will have vacation this and the coming week so unfortunately might have longer answer/fix/merge times.

alyst:
@robertfeldt Looks nice! I have left some inline comments.

multidis:
> Please check the proposed top-level use cases shown here:

Optional keyword arguments are more convenient IMO, associative array in the input seem to complicate bboptimize calls (especially with the upcoming 0.3 -> 0.4 transition).

As far as the results-object, having a possibility to restart the optimization from it would be helpful. As long as the results object is a class instance and not a closure (so that it could be saved to a JLD file).

robertfeldt:
@multidis yes, saving and restarting is great. I added a test case for it and also updated based on your, @alyst and Matthieu's feedback. Would be great if you can check again that it is in line with expectations.

@alyst , we should introduce a separate OptimizationTrace type for saving the fitness progress and related data. It has really nothing to do with the archive. When I lift in the Borg multi-objective EA code it uses an Archive actively in parallel to a normal Population and I want to stick to its nomenclature. So the Archive is just saving the best solutions found and no history per se.

robertfeldt:
So pls check

https://github.com/robertfeldt/BlackBoxOptim.jl/blob/master/design/design_by_testcode/test_bboptimize_toplevel_interface.jl

again.

multidis:
The restart functionality seems quite straightforward, very helpful addition to the package! My understanding is that since the controller type is not immutable, termination settings may be adjusted prior to the restart if needed, correct?

Looking ahead to the optimization trace history object: would it be possible to continue writing to the same object after restart? Post-optimization analysis (search vs. objective space robustness etc.) benefits from having a complete function evaluation history over the entire run. @robertfeldt you mentioned plans for multi-objective methods, in that case it would be even more useful (in some applications) for comparing parts of the Pareto front with respect to selected metrics computed through interpolation over the parameters -> objectives set acquired during the optimization run. Sorry if I am deviating too far from the current PR issue. 

One minor suggestion for the tests in the view of significant code changes: perhaps adding a test when `SearchRange` differs between search space dimensions, i.e. when an array of tuples is supplied.

robertfeldt:
I've changed the top-level interface and the first 3 new tests we discussed now works.

@alyst I had some troubles understanding the value of the Parameters typealias and the EMPTY_PARAMS const so went with a simpler solution for now. Please comment if I messed up or missed the value your construct and interface provided. We can always introduce again.

robertfeldt:
@alyst @multidis @matthieugomez 

The 3 implemented tests are here:

https://github.com/robertfeldt/BlackBoxOptim.jl/blob/master/test/test_toplevel_bboptimize.jl

Thanks for any feedback and sorry for inconvenience by changing the top-level API. Might be a few rounds of testing etc before it settles down.

robertfeldt:
BTW, sorry for doing these changes on master. Meant to use a branch but had mixed up with some other changes and decided to go ahead...

alyst:
The idea was that ```Parameters```` (````typealias Parameters Associative{Symbol,Any}```) should be the [one and only] type declaration for "parameter dictionary" arguments in BlackBoxOptim API, because it enforces the keys to be symbols. But since BlackBoxOptim should support 0.3 dicts syntax (```{:a => 3...}```) resulting in ```Dict{Any,Any}``` I had to change some ```Parameters``` into ```Associative```. Still I think it would be nice to use ```Parameters``` everywhere once 0.3 compatibility would be dropped.

```EMPTY_PARAMS``` is related. It was supposed to be  just a placeholder for the bulky ```@compat(Dict{Symbol,Any}())``` expression. But now I realize that it's not how it should be done. The proper way is to use ```empty_params() =  @compat(Dict{Symbol,Any}())```, because parameters could be modified inside the function.

**Comment by robertfeldt** [comment]
A simple to understand interface is as important IMHO. So let's revisit when we drop 0.3.

Also EMPTY_PARAMS is/was a type rather than instance of type which made it hard to understand.

**Comment by robertfeldt** [comment]
Here's its def which confused me:

```
const EMPTY_PARAMS = @compat(Dict{Symbol, Any})
```

alyst:
Oh, I have meant ```@compat(Dict{Symbol, Any}())```.

alyst:
I haven't used the updated interface yet, but I think it's much cleaner now. Thank you!

PR #23, opened 2015-07-18
Adaptation for Genetic Operators

alyst:
The PR adds ```adjust!(GeneticOperator)``` function to provide the mechanism for operator parameters adaptation. It also implements 2 adapting operators:
 * ```FAGeneticOperatorsMixture``` -- a random mixture of genetic operators that uses ```FrequencyAdapter``` to adjust the rates of each operator in the mix.
 * ```AdaptiveDiffEvoRandBin{N}``` -- ADE crossover operation

```adjust!()``` support is added to ```DiffEvoOpt```. Actually, this optimizer now supports any genetic operator. All DE/ADE-specific logic is moved into DE crossover operator classes, so maybe it could be renamed into ```EvolutionaryOptimizer``` for clarity.

Technically, ```adjust!()``` is called from ```tell!()``` for those candidates that are accepted into the population. Because ```tell!()``` doesn't know what operation was applied to the trial individual, ```op``` and ```tag``` fields were added into ```Candidate```. The first is the reference to the genetic operator applied, the 2nd stores some operator-specific information (for ```FAGeneticOperatorsMixture``` it is the index of the genetic operator from the mixture that was actually applied). If the trial individual is less fit that original, ```adjust!(op, tag)``` is still called, but with the original individual and ```new_fitness``` equal to ```old_fitness```.


alyst:
Thanks!

robertfeldt:
@alyst seems we have been getting build failures on Travis on nightly for some time. Mostly related to FactCheck complaining it seems. Any ideas?

alyst:
@robertfeldt It's due to JuliaLang/julia@5102ab28cf. Basically, it requires fixing every `@fact`, where LHS contains some operators that now have lower priority than `=>`.
I've fixed it in my local tree, but there's another issue with serializing OptController. I'll submit the PR, when it would be fixed.

PR #28, opened 2015-07-26
Fixes for recent Julia v0.4 master

alyst:
Recently, the priority of `=>` operator was changed (JuliaLang/julia@5102ab28cf), as a result, some `@fact` checks become invalid.


alyst:
So now the tests still fail, but for another reason. There's some error with `DictChain` deserialization. Interestingly, it is ok for 0.3, so it could be upstream regression. I'll try to assemble a minimal test a submit a bug report.

PR #30, opened 2015-08-17
Misc enhancements

alyst:
Some minor enhancements to the top level interface:
 * add `lastrun()` and `problem()` functions to `OptController`
 * warn upon Ctrl-C interruption
 * the result of population methods contains the whole population
 * fix v0.4 warnings

I've enabled package precompilation on v0.4 to reduce loading times. So far it looks good.

Also I've updated my scripts to the new top level interface of `BlackBoxOptim` and they are running fine, so I believe it should be sufficiently stable and the "**NOTE!**" could be removed from the readme.


alyst:
Failed nightly test is a false alarm: it's xNES again. Probably we have to tweak parameters to make it pass more often or use the "successful" RNG seed (but that is architecture-dependent).

alyst:
@robertfeldt thanks!

Issue #26, opened 2015-07-23
Invvestigate why xNES and SeparableNES often prematurely converge


robertfeldt:
Not seeing the expected performance from these two optimizers. Maybe some bug has crept in? Investigate.

Issue #19, opened 2015-07-11
xNES and SeparableNES should embed in bounding box


robertfeldt:
Check and make sure they do. Think it is not missing.

PR #31, opened 2015-08-20
Enhance NES

alyst:
Implement projecting NES candidate points into the search space (#19), specifying initial point, some linalg optimizations, exclude function compilation from timing in smoketest that frequently leads to xNES test failure (is it related to #26?).

### DIFF
--- a/src/BlackBoxOptim.jl
+++ b/src/BlackBoxOptim.jl
@@ -56,7 +56,7 @@ export  Optimizer, AskTellOptimizer, SteppingOptimizer, PopulationOptimizer,
         ParamBounds, Individual, SearchSpace, FixedDimensionSearchSpace, ContinuousSearchSpace,
         RangePerDimSearchSpace, symmetric_search_space,
         numdims, mins, maxs, deltas, ranges, range_for_dim, diameters,
-        rand_individual, rand_individuals, isinspace, rand_individuals_lhs,
+        rand_individual, rand_individuals, rand_individuals_lhs,
 
         # Population
         FitPopulation,
--- a/src/bboptimize.jl
+++ b/src/bboptimize.jl
@@ -7,7 +7,7 @@ end
 #   any-dimensional problem and a number of dimensions as a parameter
 function setup_problem(family::FunctionBasedProblemFamily, parameters::Parameters)
   # If an anydim problem was given the dimension param must have been specified.
-  if params[:NumDimensions] == :NotSpecified
+  if parameters[:NumDimensions] == :NotSpecified
     throw(ArgumentError("You MUST specify NumDimensions= when a problem family is given"))
   end
   problem = fixed_dim_problem(family, parameters[:NumDimensions])
--- a/src/direct_search_with_probabilistic_descent.jl
+++ b/src/direct_search_with_probabilistic_descent.jl
@@ -14,7 +14,7 @@ function sample_unit_hypersphere(n, num = 1)
   broadcast(*, sqrootsums, X)
 end
 
-type RandomDirectionGen <: DirectionGenerator
+immutable RandomDirectionGen <: DirectionGenerator
   numDimensions::Int
   numDirections::Int
 end
@@ -25,7 +25,7 @@ end
 
 # A MirroredRandomDirectionGen generates half of the directions randomly and then
 # mirrors by negating them.
-type MirroredRandomDirectionGen <: DirectionGenerator
+immutable MirroredRandomDirectionGen <: DirectionGenerator
   numDimensions::Int
   numDirections::Int
 
--- a/src/evaluator.jl
+++ b/src/evaluator.jl
@@ -74,6 +74,8 @@ type Candidate{F}
         new(params, index, fitness, op, tag)
 end
 
+fitness(cand::Candidate) = cand.fitness
+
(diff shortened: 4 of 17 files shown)


robertfeldt:
Thanks Alexey this looks very promising. On 0.3.11 I get a problem though:

```
ERROR: Incorrect usage of @fact: $(Expr(:-->, :(numdims(p1)), 2))
 in include at /usr/local/Cellar/julia/0.3.11/lib/julia/sys.dylib
 in include_from_node1 at /usr/local/Cellar/julia/0.3.11/lib/julia/sys.dylib
 in anonymous at no file:32
 in include at /usr/local/Cellar/julia/0.3.11/lib/julia/sys.dylib
 in process_options at /usr/local/Cellar/julia/0.3.11/lib/julia/sys.dylib
 in _start at /usr/local/Cellar/julia/0.3.11/lib/julia/sys.dylib
while loading /Users/feldt/tmp/bbo/BlackBoxOptim.jl/test/test_population.jl, in expression starting on line 55
while loading /Users/feldt/tmp/bbo/BlackBoxOptim.jl/test/runtests.jl, in expression starting on line 31
````

Any clean way we can have this work on both 0.3 and 0.4 for the time being?

**Comment by alyst** [comment]
I've just tried it on 0.3.11 with the latest packages, and it worked for me. `FactCheck` introduced `-->` not too long ago (and it should work for both 0.3 and 0.4), maybe you have an older version?

**Comment by robertfeldt** [comment]
Ah sorry, have several laptops and had missed updating on one. Looks good.

**Comment by robertfeldt** [comment]
Sorry this is taking some time but when benchmarking I sometimes run into an assert violation and it's hard to debug and seem rare:

````
ERROR: assertion failed: l <= target[i] <= u
 in apply! at /Users/feldt/tmp/bbo/BlackBoxOptim.jl/src/genetic_operators/embedding/random_bound.jl:27
 in step! at /Users/feldt/tmp/bbo/BlackBoxOptim.jl/src/generating_set_search.jl:121
 in run! at /Users/feldt/tmp/bbo/BlackBoxOptim.jl/src/opt_controller.jl:168
 in run! at /Users/feldt/tmp/bbo/BlackBoxOptim.jl/src/opt_controller.jl:277
 in bboptimize at /Users/feldt/tmp/bbo/BlackBoxOptim.jl/src/bboptimize.jl:49
 in bboptimize at /Users/feldt/tmp/bbo/BlackBoxOptim.jl/src/bboptimize.jl:48
 in fitness_for_opt at /Users/feldt/tmp/bbo/BlackBoxOptim.jl/examples/benchmarking/compare_optimizers.jl:206
 in compare_optimizers_to_benchmarks at /Users/feldt/tmp/bbo/BlackBoxOptim.jl/examples/benchmarking/compare_optimizers.jl:369
 in main at /Users/feldt/tmp/bbo/BlackBoxOptim.jl/examples/benchmarking/compare_optimizers.jl:124
 in include at /usr/local/Cellar/julia/0.3.11/lib/julia/sys.dylib
 in process_options at /usr/local/Cellar/julia/0.3.11/lib/julia/sys.dylib
 in _start at /usr/local/Cellar/julia/0.3.11/lib/julia/sys.dylib
while loading /Users/feldt/tmp/bbo/BlackBoxOptim.jl/examples/benchmarking/compare_optimizers.jl, in expression starting on line 52
`````

**Comment by robertfeldt** [comment]
Very strange it happened again and for the same problem:

````
Schwefel1.2, n = 50, optimizer = generating_set_search
````
My only theory is that the optima for this problem is very close to one of the borders and thus is more likely to surface. But it seems strange it would then happen on for dimension 50 and not for lower dimension version of the same problem. Hmm...

**Comment by alyst** [comment]
I guess it is a result of arithmetic error, `target[i] == 1.00000001` or something like that, which maybe happens is the population has converged to the optimum that lies on the border. 
This should help debugging
```julia
    @assert l <= target[i] <= u "target[$i] = $(target[i]) is out of ($l, $u)"
```
If it's the case, we can explicitly use `clamp()`.

robertfeldt:
Yes, some interesting info:

```
Schwefel1.2, n = 50, optimizer = generating_set_search, run 1 of 5 (79.01% of total runs)
Starting optimization with optimizer GeneratingSetSearcher(ConstantDirectionGen)
0.00 secs, 1 evals, 0 steps, fitness=3128561.656763172
ERROR: assertion failed: target[1] = NaN is out of (-100.0, 100.0)
 in error at error.jl:21
 in apply! at /Users/feldt/tmp/bbo/BlackBoxOptim.jl/src/genetic_operators/embedding/random_bound.jl:27
```

alyst:
I was because `GSS.step_size` was increasing until it reached `Inf`. I have set the default max step size to the maximal finite value, it looks like the issue is gone.

robertfeldt:
Ah, nice catch, thanks. Sorry it took some time but I'm traveling and currently in US. Will run some benchmarks now and then merge.

robertfeldt:
Hmm, the NES opt performance problems are not addressed by this change so still needs to be investigated. Since not many ppl use NES right now we skip this for now...

alyst:
Thanks! With this PR the number of function evals per second could decrease as every point could be projected, but I guess it is unrelated to #26.
For which problems the performance issues of NES are most apparent?

PR #34, opened 2015-09-17
add ParallelEvaluator

alyst:
Implements `ParallelEvaluator <: Evaluator` that distributes the calculation of candidates' fitnesses across the worker processes. Should be beneficial for NES family of optimizers that regenerate the population at each step.

It does require channels from Julia 0.4, but unlike `ParallelOptimizer` from #25 communication between the processes is much more straighforward (it's very similar to `pmap()`, except that the `problem` object needs to be sent to the workers only once at `ParallelEvaluator` creation).

To use it, simply add `Workers=` parameter to `bbsetup/bboptimize(problem)` call. E.g.
```julia
addprocs(2)
bbsetup(problem, Workers=workers())
```

### DIFF
--- a/src/BlackBoxOptim.jl
+++ b/src/BlackBoxOptim.jl
@@ -4,6 +4,12 @@ module BlackBoxOptim
 
 using Distributions, StatsBase, Compat
 
+# selectively enable parallel evaluation:
+# - Julia v0.3: no RemoteChannel
+# + Julia v0.4
+# - Julia v0.5: incompatible changes in parallel API
+enable_parallel_methods = VERSION >= v"0.4.0" && VERSION < v"0.5.0-"
+
 export  Optimizer, AskTellOptimizer, SteppingOptimizer, PopulationOptimizer,
         bboptimize, bbsetup, compare_optimizers,
 
@@ -110,7 +116,9 @@ include(joinpath("problems", "all_problems.jl"))
 include(joinpath("problems", "problem_family.jl"))
 
 include("evaluator.jl")
+if enable_parallel_methods
 include("parallel_evaluator.jl")
+end
 
 function setup!(o::SteppingOptimizer)
   # Do nothing, override if you need to setup prior to the optimization loop
--- a/src/BlackBoxOptim.jl
+++ b/src/BlackBoxOptim.jl
@@ -110,6 +110,7 @@ include(joinpath("problems", "all_problems.jl"))
 include(joinpath("problems", "problem_family.jl"))
 
 include("evaluator.jl")
+include("parallel_evaluator.jl")
 
 function setup!(o::SteppingOptimizer)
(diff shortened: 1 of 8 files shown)


robertfeldt:
Thanks, looks great. Can we set things up so this is only activated on 0.4 and later and that its tests are only run if on 0.4? Or maybe we just wait until 0.4 is officially released and we require it. What do you think?

alyst:
Since the 0.4 release is very near, I think it would be easier to just require 0.4 at some point, then merge this PR. Last week Julia 0.4rc2 was released, it fixes some stability issues related to interprocess communication, so in principle we can already switch the master to 0.4 and tag the new version after the official release.
Meanwhile METADATA.jl points to BlackBoxOptim v0.0.1 that supports 0.3, and as long as the new version is not tagged, 0.3 users would not be affected.

alyst:
I've rebased the PR to the master. It builds on 0.4, fails on 0.3 (the PR requires channels, so it would not work 0.3), fails on nightly (the core devs are changing the parallel API, but I don't think it makes sense to update the PR now. I expect there would be more changes in the 0.5 development cycle).
So accepting this PR requires dropping 0.3 compatibility.
I was using it quite extensively for several different problems in a cluster environment, and it was looking stable.

robertfeldt:
Sounds good. I'm ready to require 0.4. Will check and run on my side and then merge. @alyst , do you recommend we also exclude nightly from the travis build/testing? Disturbing not to have a travis pass.

alyst:
IMO nightlies should be kept. I've added the commit disabling ParallelEvaulator on 0.3 and 0.5. Unfortunately, it didn't worked out:
 * on 0.3 it's ctor compatibility issue
 * on 0.5 it's some strange error in sorting along the dimension in the latin cube. It might be the problem with Julia nightlies that would be resolved soon.

As for 0.3, I think it should be dropped. `BlackBoxOptim` extensively uses type parameters, and support for type parameters was much enhanced in 0.4, so maintaining 0.3 compatibility would mean writing less elegant code.

robertfeldt:
Ok, I've tried this and it looks good. I think we should also add an example file that runs an opt first with only one worker process and then with several, to show people how it can be used and what the benefits are. Will see if I have time to add one, if not I'd appreciate one from you. :)

robertfeldt:
Hmm, I have problems ensuring all code is loaded on all workers and get something like:

```
ERROR: LoadError: On worker 4:
UndefVarError: ParallelEvaluatorWorker not defined
 in anonymous at /Users/feldt/tmp/bbo/BlackBoxOptim.jl/src/parallel_evaluator.jl:117
...
```

after trying to addprocs() and then setup and run an optimization with multiple Workers. Maybe I need a small example from you anyway... Can you please check and do a PR on this file:

https://github.com/robertfeldt/BlackBoxOptim.jl/blob/master/examples/rosenbrock_parallel.jl

Issue #36, opened 2015-09-30
Non-population optimizer test fails intermittently


tkelman:
I've been running PackageEvaluator a bunch over the past few weeks and this package has been intermittently failing. Something about the tolerances or random inputs, or an actual bug?
```
Top-level interface
  > run a simple optimization
    > using bboptimize() with mostly defaults
    > using bbsetup()/bboptimize() with mostly defaults
    > using non-population optimizer
      Failure :: (line:-1) :: using non-population optimizer :: fact was false
        Expression: best_fitness(res) --> less_than(1.0)
          Expected: 3.5080192708609457 < 1.0
    > using population optimizer
  > continue running an optimization after it finished
  > continue running an optimization after serializing to disc
Out of 46 total facts:
  Verified: 45
  Failed:   1
```

robertfeldt:
Should only be too tight tolerances so I upped the iteration count and hopefully this should be very unlikely. Get back to me if not.

tkelman:
Appreciate it! I'll have to check in a non-pkgeval run to see if I can reproduce the issue and if checkout to master makes it go away. PkgEval runs on the most recently tagged version.

robertfeldt:
Ok, I run a number of times here without reproducing it now so should be very unlikely. Since most of our testing is "random" (i.e. starts from random seeds) it will sometimes happen though; this is the nature of testing non-deterministic opt algs imho since fixing seeds would not provide as diverse testing. So I'd rather tune the tolerances...

tkelman:
I got another one happening at a different place, probably also intermittent?
```
bboptimize smoketest
  > testing resampling_memetic_search method to ensure it works
  > testing de_rand_2_bin_radiuslimited method to ensure it works
  > testing de_rand_2_bin method to ensure it works
  > testing de_rand_1_bin_radiuslimited method to ensure it works
  > testing adaptive_de_rand_1_bin_radiuslimited method to ensure it works
  > testing adaptive_de_rand_1_bin method to ensure it works
  > testing separable_nes method to ensure it works
  > testing dxnes method to ensure it works
ERROR: LoadError: LoadError: AssertionError: target[2]=NaN is out of [-2.0, 2.0]
 in apply! at /home/vagrant/.julia/v0.4/BlackBoxOptim/src/genetic_operators/embedding/random_bound.jl:27
[...]

PR #37, opened 2015-10-27
DX-NES fixes

alyst:
Some small NES enhancements (`:ini_x` checks and introduction of `:ini_lnB` parameter) plus an attempt to fix failing DX-NES tests (#36):
 * step size (`sigma`) is limited by `max_sigma` (defaults to 1E10) to avoid floating point overflows
 * scale distance (in the original paper the algorithm pseudocode has some differences from the description in the text; the fix makes it as discussed in the text)
 * change how the moving threshold is calculated. See below the explanations (keeping in mind that I'm not a population methods expert).

For my optimization problems (that have >100 dimensions) the last point was critical. In fact, the evolutionary path and moving threshold is not a DX-NES novelty, it came from CMA-ES. It's assumed there that if the population is not moving as the whole, the evolutionary path should be a random N-dimensional vector and so the moving threshold is `E|path|`, `|path| ~ Chi(N)` (in fact, `E|path|^2 = N`, `|path|^2 ~ ChiSqr(N)` is even simpler). But, actually, there should be at least 3 possibilities:
 * `deterministic movement` The population lies "on the same slope" and is moving as the whole towards minimum. Hence `u*Z` is big, and evol. path grows as at every iteration the population moves in the same direction. The step (sigma) should be increased to facilitate convergence.
 * `refining minimum` The population covers minimum (hence "natural gradient" `u*Z` is small). The step should be decreased to locate the minimum more precisely.
 * `chaotic movement`. The population is on the slope (so `u*Z` is large), but the step size (`sigma`) is too large, and the new population at next iteration misses the minimum, i.e. the method is diverging. Logically, the step size should be reduced. But in this situation `|path| ~ Chi(N)` is no longer true, because `u*Z` is large. In fact, evolutionary path is a simple autoregressive process, not `~N(0,I)`, and the moving threshold should be set accordingly (see 8649447), otherwise the method thinks it's in `deterministic movement` state and increases `sigma`, which leads to divergence.

### DIFF
--- a/src/dx_nes.jl
+++ b/src/dx_nes.jl
@@ -8,7 +8,7 @@ type DXNESOpt{F,E<:EmbeddingOperator} <: ExponentialNaturalEvolutionStrategyOpt
   B_learnrate::Float64
   sigma_learnrate::Float64
   max_sigma::Float64
-  moving_threshold::Float64       # threshold of evolutionary path movement
+  moving_threshold::Float64       # threshold of |evolutionary path| to detect movement
   evol_discount::Float64
   evol_Zscale::Float64
   uZscale::Float64                # distance scale for the weights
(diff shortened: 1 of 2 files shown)


robertfeldt:
Sounds good. I'm no expert on CMA-ES or NES myself but this sounds like good improvements. There will always be a tension for us between implementing things as originally described (with details often missing ;)) or based on what actually works when we test it. We need to find a pragmatic way forward in each case. Sometimes it might be better to keep the original (close to orig paper) implementation and do a small "diff" implementation of what we see work best. But if not then we go for what works... :)

PR #40, opened 2015-11-28
Fix the parallel Rosenbrock example

alyst:
This fix works for me. `rosenbrock()` should be defined `@everywhere`, OTOH `@everywhere` is not required for `using`. I guess there was also a problem of first using BlackBoxOptim on the master, and then adding the workers, so the workers are initialized in the very beginning.

Actually for me the performance of this example for the parallel method is 3 times lower than serial. Perhaps the function is too simple and there's more communication slowdowns than gain from parallel evaluation. Probably, instead of giving candidates to worker one-by-one, it should be given a population subset at once.

### DIFF
--- a/examples/rosenbrock_parallel.jl
+++ b/examples/rosenbrock_parallel.jl
@@ -1,20 +1,22 @@
+# Now add 2 procs that can exec in parallel (obviously it depends on your CPU
+# what you actually gain from this though)
+addprocs(2)
+
+# note that BlackBoxOptim gets automatically used on all workers
 using BlackBoxOptim
 
-function rosenbrock(x)
+# define the function to optimize on all workers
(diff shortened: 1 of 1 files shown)


robertfeldt:
Thanks, yes I see it is obvious now. Maybe adding a delay in the rosenbrock func makes for a bit more realistic example but anyway.

robertfeldt:
Strange I don't seem to get any performance difference between the serial and parallel runs. Will play around with the params but this seems strange. I'm currently on a 2-core machine only but expected to see some difference.

robertfeldt:
I had a stale cache file. Fixed now and I get speed-ups from 1.3-1.9x single-core performance depending on how much delay I add in the func to be optimized. Nice work @alyst , thanks.

PR #42, opened 2016-02-04
Convert comments to v0.4 docstrings, drop v0.3 support.

alyst:
The PR converts all the type/method documentation comments into docstrings that were introduced in v0.4. With this change it's possible to retrieve (some) documentation for `BlackBoxOptim` within interactive Julia session.

The support for Julia v0.3 is dropped, because it doesn't support the docstrings.
There's also some minor edits to the comments for clarity.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
   - linux
   - osx
 julia:
-  - 0.3
   - 0.4
   - nightly
 notifications:
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
-julia 0.3
+julia 0.4
 FactCheck
 StatsBase
 Distributions
--- a/examples/benchmarking/compare_optimizers.jl
+++ b/examples/benchmarking/compare_optimizers.jl
@@ -130,7 +130,7 @@ function main(args)
   close(logfilehandle)
 end
 
-ProblemSets = @compat Dict{ASCIIString,Any}(
+ProblemSets = Dict{ASCIIString,Any}(
   "easy" => [
     # ProblemName, NumDims, PopSize, MaxFevals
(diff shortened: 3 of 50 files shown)


robertfeldt:
Sorry for this taking some time. I'm on part time parental leave so haven't had much time. The failed check here is the "normal" one, right?

alyst:
No problem!

I think the failure is the "normal" one, there were some big anonymous function changes in 0.5 at the time of the PR, so many packages were not compatible with that change yet.

But I'll update the docstrings format according to the new guidelines and we'll see if it builds now.

alyst:
I've updated the docstrings, also continued with the removal of 0.3 legacy (`@compat`, old-style array ctors, replacing `Array{T,N}` with `Vector/Matrix{T}, cleaning up `Dict{}` usage in the context of method parameters).

Unfortunately, the package still fails in v0.5 due to julialang/julia#14919 (I want the `FitnessScheme` to behave like a function that defines the ordering of fitness values, then it's very convenient to pass it to sorting methods and no anonymous function performance penalty is introduced). I hope that issue would be fixed later in v0.5 development cycle, meanwhile I think it's ok to have it as is, since we are not really using v0.5 yet.

alyst:
I had yet another attempt to make v0.5 working, but it failed again, this time because of JuliaLang/julia#15068 :)

alyst:
Should pass v0.5 tests now (there's Travis issue with the v0.5 image on Mac, but I think it's not related to the package). Because now v0.4 uses anonymous functions instead of `Base.call()` for fitness sorting, it would be nice to check whether the performance is the same as before.

robertfeldt:
Sounds great, I'll run perf tests and check it out this weekend.

Issue #41, opened 2016-01-27
Multi-objective optimization


alyst:
Currently I'm considering whether multicriterion optimization (MO) can improve the convergence for my problems (Bayesian probability-like functions, i.e. prior probability is one criterion, likelihood is another).

Would it make sense to extend `BlackBoxOptim` to multicriterion problems? If yes, what would be the best strategy of implementing MO? What are the most efficient evolutionary methods for MO? Should MO be constrained to 2-criteria case (as then the Pareto frontier is easier to manipulate)?

robertfeldt:
Do you mean multi-objective, i.e. multiple goals that needs to be traded off, typically with a Pareto Front / Hypervolume approach? If so the answer is definitely yes, we use that in our research and have planned to include it properly in BBO at some point. I consider the state-of-the-art to be the Borg MOE of Hadka et al and have planned to prioritize/focus the design around that, see for example:

Hadka, David, and Patrick Reed. "Borg: An auto-adaptive many-objective evolutionary computing framework." Evolutionary Computation 21.2 (2013): 231-259. 
https://scholar.google.se/scholar?hl=sv&q=Borg%3A+An+Auto-Adaptive+Many-Objective+Evolutionary+Computing+Framework&btnG=

Currently I do not have a lot of time for this though so if you are in a hurry maybe you can consider starting on the implementation? Or we can set up a way to collab on it?

robertfeldt:
And no, I think we should not constrain to 2 objectives. Many of the recent many-objective algorithms of Hadka et al, Deb et al, Yao et al and others can scale from 2 objectives up to 15-30 of them quite gracefully. The older NSGA-II is known to only work well for quite few objectives so I would rather not start there. IMHO, the Borg approach is one of the most elegant and effective and can dynamically trade off between different operators as the search needs them.

Some of Deb et al's recent work even shows that there are many-objective algorithms that scale down to even work well in single-objective settings.

alyst:
Yes, I've meant multi-objective optimization. Thank you for the references! At the moment I don't have resources to start implementing it myself either, I just wanted to better understand how big and relevant would be the effort. Of course, if somebody would start implementing MOO, I will be very much interested in contributing/testing etc.

robertfeldt:
Ok, I plan to start this work some time in March-April 2016. If you have time to wait for that it would be great if you can then help out, test, and refine it. Thanks.

alyst:
That sounds great!

PR #8, opened 2015-05-05
Avoiding merge dispatch conflicts

multidis:
Resolving `Base.merge`-related issues involving the `Parameter` type.

Rather than exploring all possible conflicts with the `merge` dispatch, it seemed more straightforward to just use a different function name for combining parameter dictionaries. Tests in `test_parameters.jl` all pass, but @robertfeldt please re-run the whole test setup with your custom `Rakefile` (have not read into that).

Thanks @MichaelHatherly for [debugging](https://github.com/robertfeldt/BlackBoxOptim.jl/issues/7#issuecomment-98963828) the problem.

PR #43, opened 2016-03-09
[WIP] Support Multi-Objective optimization + Borg MOEA

alyst:
This is the initial version of the multi-objective optimization support + the initial implementation of the Borg MOEA (see #41). Should be applied on top of #42.

Changes the core components to support multi-objective optimization. Some worth mentioning:
* the fitness defined in the `OptimizationProblem` could be different from the fitness stored by the `Archive` (needed for fitness epsilon-indexing), so `Evaluator` supports these 2 fitness types
* "transient" members for `FitPopulation`. Transient individuals are not part of the population (ignored when selecting parents from the population), but could be indexed as any other members. This is needed for Borg recombination, where one of the parents should be from archive.
* vector-based fitness draft was replaced by tuple-based fitness (`NTuple{N,F}`). This should reduce the GC overhead as no heap memory is allocated for vector contents. Also the generated code could be potentially more efficient since `N` is known at the compile time.

Added `TournamentSelector` individuals selection operator.

Added `TupleFitnessScheme` abstract class and its children:
 * `ParetoFitnessScheme`
 * `EpsDominanceFitnessScheme`
 * `EpsBoxDominanceFitnessScheme`

Each scheme could be used in minimization (component-wise) or maximization modes.

Added `EpsBoxArchive` as it was described in Hadka & Reed 2013 Borg paper.

Added Borg MOEA methods.

TODO
- [ ] `TournamentSelector` tests
- [ ] `SimulatedBinaryCrossover` operator + tests
- [ ] `ParentCentricCrossover` operator + tests
- [ ] `UnimodalNormalDistributionCrossover` operator + tests
- [ ] `SimplexCrossover` operator + tests
- [ ] more multi-objective test problems (at the moment only `Schaffer1()`)
- [ ] support for `ParallelEvaluator`


robertfeldt:
Thanks for this, very nice to get this started. The failures don't look too serious, tolerance issue? I'm holding off on merging this until we've tested it more...

alyst:
Yes, the failure is due to the tolerance: the Pareto frontier for `schaffer1()` (which is the only multiobjective test I have so far) almost coincides with tangent line at the best solution.
ATM I'm adding a few more crossover operators to make sure that the adaptive recombination works. If you want to merge it early and then polish it on the master, maybe then it would be the time.

Definitely, we need to have a comprehensive set of the test multiobjective problems. But just to cover more potential issues with `TupleFitness`/`EpsBoxArchive`, we need to add at least one >2-objective problem into this PR. Do you have some in mind?

robertfeldt:
ZDT would be the "iconic" one but any of the ones in:

http://dces.essex.ac.uk/staff/zhang/MOEAcompetition/cec09testproblem0904.pdf.pdf

is a good starting point, I guess.

robertfeldt:
Actually it's probably better to go with some other than ZDT, here is more info:

http://dces.essex.ac.uk/staff/zhang/moeacompetition09.htm

alyst:
I've updated the PR:
* `FunctionBasedProblem` handling of bounded/unbounded problems was simplified. For multicriterion problems it's now possible to specify `opt_value=Hypersurface(f, tss)`, where `f(t)` is a parametric definition of the known Pareto Frontier, `t in tss`.
* Currently `problem/multi_objective.jl` defines `Schaffer1` family of 2-objective problems and `CEC09` 3-objective Unconstrained Problem `#8`.
* `OptimizationResult` was simiplified, for multi-objective problems it supports `pareto_frontier(opt_res)` (or should it be called `pareto_set()`?; actually it returns the structure with both the fitness and the corresponding point in the search space)
* To test the MO method convergence to the known Pareto Frontier, `IGD(ref_surf, solution)` is calculated, as proposed in CEC09 tests doc. However, AFAIU the original `IGD()` only checks whether `ref_surf` lies within the neighbourhood of `solution`. There are two problems: 1) it's not checked whether the opposite is true, and `solution` lies within `ref_surf`; 2) `ref_surf` points are epsilon-indexed, but not checked for epsilon-dominance, so it's possible that dominated point from `ref_surf` would be far away from any point of the correct solution (which contains only epsilon-nondominated points).  To overcome this, two checks are done: whether the`solution` lies within `ref_surf` and whether the nondominated set of `ref_surf` points lies within `solution`. Most probably, it's a very well known problem and it's already properly addressed, I didn't do an extensive research.
* SPX and SBX operators (without unit tests).

Obviously lots of unit tests are still missing, but at least most of the changes to the common framework are there. I would also like to add parallelization support, but not sure yet what is the most easy way; parallelization might also require some changes to the common code.

robertfeldt:
Wow, this looks good. A few questions:

* Do you support epsilon in the EpsBoxArchive to be a vector of different epsilon-values per fitness dimension? This can be quite important if they have different "scales".

* How stable is this? Should we integrate with base now or wait? Even if there are a few bugs left I'm ok with integrating as long as it seems stable enough to run on a few simple MO problems.

alyst:
* per-objective epsilons are not yet supported, but it should be easy. I'll try to do it for the next update
* I think the refactoring of the core components is rather stable, it shouldn't introduce instabilities to the existing one-objective methods. The MOEA code is not well tested, I expect it would be tweaked a lot, but so far it's not crashing on the simple problems I have tried.

I'll try to apply Borg to my optimization problem and address all the corresponding issues in the next update. After that, if you want to work on the code, maybe the minimal thing before merging would be to run the benchmark to check that one-objective methods have not regressed.

robertfeldt:
Sounds good. I'll hold off a little bit more then, before we integrate.

Note that there is a new rake command "rake bench" which will run the (single-objective) benchmarking and compare to latest, saved benchmark. It will also use Benjamini-Hochberg correction for the multiple comparisons so the final count of significant differences is > 0 something is likely to have affected the core components.

robertfeldt:
BTW, would be great with short and simple multi-objective example in examples subdir. Longer-term I would also like to have the shortest possible MO example for the README since it's such a fundamental feature we should highlight it early for newcomers.

alyst:
I have fixed a couple of issues and allowed per-objective epsilon. So it should be in better shape for merging.

robertfeldt:
Thanks. Tests pass and I'm running benchmarking now. One thing I noticed is that fitness values are now printed as strings which looks weird:

```
Best candidate found: 
...

Fitness: "6958.636392643"
```

alyst:
I've introduced `format_fitness()` method. It's used for the final output and also for logging, and I think I have taken the formatting string from the old logging code. How you would like the fitness to be printed?

I've also worked a little bit on the crossover operators and fixed some bugs. I don't know how to properly unit test them, but the at least plots of the offspring distribution look like it's described in the literature (I would put the plot-generating script into the examples folder). I would submit the updated PR in a few of hours. It should not affect the old code. So if you would like to merge the PR, maybe you can wait until that update?

robertfeldt:
Ok, great I'll wait with the merge then and can fiddle with format_fitness if I see a need. The benchmarking results are fine so no effects on the old code that I can detect. Nice work.

robertfeldt:
The `fitness_format` looks fine btw, the problem is that I've used `showcompact` on that string. When I've merged I can change so it is printed as is if it is a string. Or maybe we never need `showcompact` anymore; I don't remember the reason I used it originally.

robertfeldt:
Oh, and please add the plots for the operators. That sounds interesting as a testing method... Testing non-deterministic code is hard. One technique is to fake the random number generation while doing it i.e. you would have to create a subtype to `AbstractRNG` and we'd have to use that in all `rand` calls. Longer-term we should do this anyhow since it allows repeatable experiments and sensitivity analysis etc. Anyway, it can wait for now.

alyst:
I've updated the PR. PCX and UNDX are added to the pool of crossover operators. I hope I've got the linear algebra right (esp. scaling of the parent subspace basis), otherwise the parameters could have different behaviour that in the original papers.

robertfeldt:
Thanks. I got this when running the test suite:

````
  > CEC09_UP8
    Failure :: (line:505) :: CEC09_UP8 :: fact was false
      Expression: BlackBoxOptim.IGD(BlackBoxOptim.CEC09_Unconstrained_Set[8].opt_value,pareto_frontier(res),fitness_scheme(res),Val{length(best_candidate(res))}) --> less_than(0.1)
        Expected: 0.10701014367027448 < 0.1
```

I guess we can set it to 0.15 at least for now? Have you any feeling for how performance is compared to some of the published papers using CEC09 benchmarking test suite?

alyst:
Thanks for merging!
I was not looking at the comparison with the other methods. Maybe the best way to start and check whether the implementation is correct is to compare with the original Borg MOEA run with the very same settings.

robertfeldt:
Yes, I'll see if I get some time this weekend to check some of the problems they used. I'd also be very happy if you report back if you can successfully use this for your problem. :) Thanks again, this is very useful.

robertfeldt:
@alyst do you have a short example of using a vector of epsilon-values, different ones per objective? I tried this:

```julia
schaffer1(x) = (10.0 * sumabs2(x), sumabs2(x .- 2.0))
res = bboptimize(schaffer1; Method=:borg_moea,
                FitnessScheme=ParetoFitnessScheme{2}(is_minimizing=true, ϵ = [5.0, 0.1]),
                SearchRange=(-10.0, 10.0), NumDimensions=3, ϵ=0.1,
                MaxSteps=15000, TraceInterval=1.0, TraceMode=:verbose);
````

but no luck.

robertfeldt:
Sorry forget that question, I found that epsilon is a normal param. Silly me.

PR #46, opened 2016-04-21
[RFC/WIP] Asynchronous ParallelEvaluator

alyst:
The PR changes the `ParallelEvaluator` to be asynchronous:
 1. The candidate is submitted to the queue by `async_update_fitness()`, which immediately returns the fitness calculation job Id and the job gets submitted to one of the available worker processes.
 2. The job status could be tested by the `isready()` function. Also, the candidates with the recently calculated fitness could be processed by calling `process_completed() do job_id, candidate <custom code> end` routine. (Note: if `isready(job_id)` is called and it returns true, this job would not be enumerated by a call to `process_completed()` anymore and it's up for the caller to take actions).

The old synchronous API (used by NES) is still supported.

`BorgMOEA` is updated to support asynchronous `ParallelEvaluator`: the algorithm runs on master, the new individual is generated by recombination and sent to the parallel evaluation, the further processing of the individual (updating the population and the archive) is postponed until its fitness is evaluated. This should improve the performance for the problems with computationally intensive fitness functions. I see the speed up for my problems, although it's not linear.

There are several reasons why it's RFC/WIP:
  1. Unfortunately, the current julia parallelization scheme (remote channels/remote refs implemented using the messages sent through pipes) doesn't allow for an efficient intensive communication between the parallel processes (the profiling shows that the overhead of serializing/deserializing the message and writing/reading to UV pipe stream is too high), so after trying several implementations (this journey could be tracked as I haven't squashed the commits yet) I've ended up using shared arrays for the data and status exchange. Using threads (or maybe better messaging implementation) would be much more efficient, but it's not yet mainlined to Julia.
  2. The logic of `async_update_fitness()`/`isready()`/`process_completed()` is somewhat confusing, but so far I had no better ideas given that `Future{T}` approach would create too much overhead.
  3. As the communication is shmem-based, the workers use "busy wait" to continuously poll for the new jobs. To avoid 100% CPU usage, the workers have to be killed/suspended after the optimization is finished. This is done by the new `shutdown!()` call. However, `OptController/OptRunController` interface assumes the same optimization method (and its evaluator) could be reused in several `OptRunController` runs. With the `ParallelEvaluator` it is currently not possible, because all the workers are killed by the `shutdown!()` call at the end of the first run. I see two alternatives:
  * introduce start!(Evaluator)/shutdown!(Evaluator) methods that needs to be called immediately before/after method iterations (no-ops for the normal evaluators);
  * implement lazy initialization of the workers in `ParallelEvaluator`, detect idle periods, hybernate the workers when idle (using `wait()`), resume on the new fitness evaluation request. Given the state of the parallelism in julia it's not so easy to get it working nicely.


robertfeldt:
@alyst waht is the status of this branch now? In light of Julia 1.0 soon seems to be coming can we try to unify the different parallellization branches and ideas and merge with master?

PR #52, opened 2016-08-04
WIP: Fix dep warns

ranjanan:


ranjanan:
@robertfeldt It fails on v0.5 because the package defines a `view` function too. 
```
ERROR: LoadError: LoadError: MethodError: view(::Array{Float64,2}, ::Colon) is ambiguous. Candidates:
  view(pop::Array{Float64,2}, indi_ix) at /home/travis/.julia/v0.5/BlackBoxOptim/src/population.jl:27
  view(A::AbstractArray, i::Union{AbstractArray,Colon,Real}) at subarray.jl:68
```

robertfeldt:
Thanks for this, looks great. Help me understand the issue with `view` more: do you suggest we rename our `view` or is there some alternative solution?

ranjanan:
Umm, it depends on how often you're using `view` in the package source. Is its usage small enough to just change the name and change it in a few places?

robertfeldt:
Ok, I'm trying this now. Except for the `view`ambiguity error do you also get a few more dep warnings on 0.5 like so:

```julia
WARNING: deprecated syntax "call(hc::HatCompare, ...)".
Use "(hc::HatCompare)(...)" instead.

WARNING: deprecated syntax "call(hc::HatCompare{FS}, ...)".
Use "(hc::HatCompare{FS})(...)" instead.

WARNING: deprecated syntax "Base.(:(==))".
Use "Base.:(==)" instead.
WARNING: Base.UTF8String is deprecated, use String instead.
  likely near /Users/feldt/tmp/bbo/BlackBoxOptim.jl/src/problems/single_objective.jl:181
```

and on 0.4:

```julia
    slice(AbstractArray, Union{Base.Colon, Int64, AbstractArray{T<:Any, 1}}...) at subarray.jl:39.
To fix, define
    slice(Array{Float64, 2}, Union{Base.Colon, Int64, AbstractArray{T<:Any, 1}})
before the new definition.
WARNING: New definition
    slice(Array{Float64, 2}, Any) at /Users/feldt/tmp/bbo/BlackBoxOptim.jl/src/population.jl:27
is ambiguous with:
    slice(AbstractArray, Tuple{Vararg{Union{Base.Colon, Int64, AbstractArray{T<:Any, 1}}}}) at subarray.jl:40.
To fix, define
    slice(Array{Float64, 2}, Tuple{Vararg{Union{Base.Colon, Int64, AbstractArray{T<:Any, 1}}}})
before the new definition.
```

? I'm willing to rename our `view` to avoid the ambiguity but should I merge and then we fix this on master or do you want to fix this on your branch and I'll then merge?

robertfeldt:
I find it very disturbing that the tests take 2-3 times longer to run on 0.5 than on 0.4. This is unexpected and we may need to investigate this in detail later.

ranjanan:
>I find it very disturbing that the tests take 2-3 times longer to run on 0.5 than on 0.4. This is unexpected and we may need to investigate this in detail later.

I think deprecation warnings slow down tests dramatically. So fixing this should make your tests faster.

>Except for the viewambiguity error do you also get a few more dep warnings on 0.5 like so:

I must fix those actually. I shall update this PR. 

>and on 0.4:

I'll look into this too, and let you know.

ranjanan:
@robertfeldt The latest commit fixed the PR and now tests pass on 0.5 too.

robertfeldt:
Great, thanks. This passes on both 0.5 and 0.4 without warnings for me now. 

Time for me on 0.4 is 1m13.7s while on 0.5 it is 4m1.4s so me and @alyst really need to look into this once 0.5 is stable.

I'll go ahead and merge soon, thanks again.

robertfeldt:
I updated to latest 0.5 (0.5.0-rc1+1 2016-08-05 15:23 UTC) and the time to run the test suite is now 3m11s so there seems to be quite a lot of performance variation between 0.5 versions. Thus we should revisit performance on 0.5 once it has been released. Without measuring it seems to me that one large time diff between 0.4 and 0.5 is the testing of the `FAGeneticOperatorsMixture` mutation operator. So we might start looking there.

Issue #51, opened 2016-07-25
Tests fail on v0.5


ranjanan:
```julia
 Pkg.add("BlackBoxOptim"); Pkg.checkout("BlackBoxOptim"); Pkg.test("BlackBoxOptim")
```
gives 
```
ERROR: LoadError: FactCheck finished with 2 non-successful tests.
 in exitstatus() at /Users/ranjan/.julia/v0.5/FactCheck/src/FactCheck.jl:568
 in include_from_node1(::String) at ./loading.jl:426
 in process_options(::Base.JLOptions) at ./client.jl:262
 in _start() at ./client.jl:318
while loading /Users/ranjan/.julia/v0.5/BlackBoxOptim/test/runtests.jl, in expression starting on line 44
```

robertfeldt:
Thanks, yes, this is a known problem since there have been a somewhat changing API for some new parts of the 0.5 release, right, @alyst ?

alyst:
@robertfeldt Yes, parallelization in particular. I've not tried BBO with 0.5 yet, I plan to do that after the release is out.

robertfeldt:
I actually built the latest 0.5 `Version 0.5.0-rc0+53 (2016-07-29 04:07 UTC)` and then 

````julia
Pkg.add("BlackBoxOptim"); Pkg.checkout("BlackBoxOptim"); Pkg.test("BlackBoxOptim")
````

and got:

````julia
...
INFO: BlackBoxOptim tests passed
459.801207 seconds (605 allocations: 31.203 KB)
````

**Comment by robertfeldt** [comment]
@ranjanan which Julia 0.5 version did you try on?

**Comment by ranjanan** [comment]
@robertfeldt This is my Julia version: 
```
Julia Version 0.5.0-rc0+30
Commit e717ded* (2016-07-28 00:39 UTC)
Platform Info:
  System: Darwin (x86_64-apple-darwin15.4.0)
  CPU: Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.7.1 (ORCJIT, broadwell)
```

and these are the specific errors I got: 
```
bboptimize() multi-objective methods smoketest
  > borg_moea
[...]

PR #53, opened 2016-08-17
Misc updates

alyst:
Another round of updates, mostly cleanups; the algorithms should not be affected:
* more Julia v0.5 compatibility fixes
* cleanups of the internal `Individual` and `Fitness` API
* clean up of the public API to access frontier elements (fitness and individual). No more ugly `fitness(i).orig[1]` constructs to access the method-independent fitness.
* `shutdown!(optimizer)` (required by asynchronous parallel evaluator that is not a part of this PR)
* update stability tests

The last commit makes the detection of stability issues on v0.5 (see #51) more robust. I observed that if `MaxTime` is small (0.3), the first gc might not happen during the first optimization run(s). Since the initial gc also triggers some time-consuming compilation, the result of that run(s) would be also off. It could explain the problems observed in #51 on v0.5. The updated script excludes any iteration with deviating running time from the benchmark.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ os:
   - osx
 julia:
   - 0.4
+  - 0.5
   - nightly
 matrix:
   allow_failures:
--- a/README.md
+++ b/README.md
@@ -89,14 +89,14 @@ can apply different criteria to filter the solutions.
 For example, to find the solution with the minimal first objective:
 ```julia
 pf = pareto_frontier(res)
-best_obj1, idx_obj1 = findmin(map(i -> fitness(i).orig[1], pf))
+best_obj1, idx_obj1 = findmin(map(elm -> fitness(elm)[1], pf))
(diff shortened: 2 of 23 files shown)


robertfeldt:
This looks good. Time taken on 0.5 is still disturbingly longer than on 0.4 (3m57s vs 0m57s for me right now) but it also fluctuates quite a bit between 0.5 builds so let's investigate this later.

alyst:
@robertfeldt I'm starting to migrate to v0.5, so I will soon have a look at BBOptim profiling for bottlenecks.

robertfeldt:
Ok, yes, sounds good. I'm back from vacation now and after a startup phase I will have more time for BBOptim dev later this autumn. Should I merge this now or do you want to work some more on minor updates? It looks good and clean to me.

alyst:
This code was staging in my branch for a while, so it should be safe to merge. Thank you!

PR #50, opened 2016-06-26
Add OptimizationValue option

matthieugomez:
Solves issue #49

PR #54, opened 2016-08-17
Rename OptimizationValue to OptimalFitness

alyst:
The PR makes some minor simplifications and cleanups to #50 that introduced `OptimizationValue` option. I also suggest to rename it to `OptimalFitness`, so it uses the same vocabulary as the rest of BBO API (or even to `TargetFitness`, since the specified value might not be optimal, but the method will stop as soon as the best solution reaches it).

**cc** @matthieugomez

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,8 +110,8 @@ In addition to the `Method` parameter, there are many other parameters you can c
 * `MaxFuncEvals`: How many evaluations that are allowed of the function being optimized.
 * `TraceMode`: How optimization progress should be displayed (`:silent`, `:compact`, `:verbose`). Defaults to `:compact` that outputs current number of fitness evaluations and best value each `TraceInterval` seconds.
 * `PopulationSize`: How large is the initial population for population-based optimizers? Defaults to `50`.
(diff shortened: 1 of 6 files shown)


robertfeldt:
I don't have a strong opinion here since this is not an option I use much myself. But the name change sounds good to me. Would be good to hear what @matthieugomez thinks since he might be one of few users of this feature, yet.

matthieugomez:
I did not really think about the rest of the API so Alexey is probably
right.

robertfeldt:
Ok, good we should change from `OptimizationValue` then. I prefer `TargetFitness` to `OptimalFitness` though since the former invites a more general use (stating the fitness level one thinks is sufficient even if one does not know whether it is optimal or not). Agreed? @alyst and @matthieugomez ?

alyst:
+1 for `TargetFitness`

matthieugomez:
Sure!

alyst:
@robertfeldt  oops, I think you merged the PR 1 second before I pushed the updated version with the parameter renamed :)

robertfeldt:
Yup, too quick on the trigger there, sorry... ;) Will fix in next PR...

Issue #63, opened 2017-04-22
Make BlackBoxOptim compatible with Julia 0.5 and above


Ayush-iitkgp:
Hello,
We want to provide support of BlackBoxOptim for global optimization in [DiffEqParamEstim.jl](https://github.com/JuliaDiffEq/DiffEqParamEstim.jl). But it is throwing lots of warnings in Julia 0.5 and also the error for 30 dimension Rosenbrock function (ERROR: MethodError: no method matching calculate_evol_path_params(::Int32, ::Array{Float64,1})). Is there any possibility of making it compatible for Julia v0.5 and above?

cc - @ChrisRackauckas

ChrisRackauckas:
Last tag was in July. Did you try checking out master?

Ayush-iitkgp:
`Pkg.status("BlackBoxOptim")
` 
BlackBoxOptim                 0.2.1+             master
The above error persists but the warnings are gone.

ChrisRackauckas:
Are you on 32-bit Windows? This package seems to have mistakenly used `Int64` like Optim did...

ChrisRackauckas:
Try my branch. https://github.com/robertfeldt/BlackBoxOptim.jl/pull/64 . It fixes the 32-bit issues.

robertfeldt:
I just tagged a new version so should soon be merged on main METADATA, hopefully. I'd like to step up and extend BlackBoxOptim again so please keep any ideas and code/PR's coming. Thanks for your interest.

PR #64, opened 2017-04-22
Fix improper typing

ChrisRackauckas:
There are hardcoded Int64's which will break usage on 32-bit OS's. This fixes the issue by using Int (and so 64-bit is completely unchanged).

### DIFF
--- a/examples/benchmarking/compare_optimizers.jl
+++ b/examples/benchmarking/compare_optimizers.jl
@@ -64,7 +64,7 @@ function main(args)
         help = "name of optimizer or optimizer set"
 
       "--numreps", "-n"
-        arg_type = Int64
+        arg_type = Int
         default = 10
         help = "number of repetitions per problem and optimizer"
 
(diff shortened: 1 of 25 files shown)


ChrisRackauckas:
This fixes at least the reported errors from https://github.com/robertfeldt/BlackBoxOptim.jl/issues/63

robertfeldt:
Ok, this sounds good. I want to clean up the latest tagged version anyway and get some new stuff in there. Will check this and get back to you.

robertfeldt:
This is against master, right?

ChrisRackauckas:
Yes.

robertfeldt:
Ok, it looks good but one test fails on 64-bit machine since shot() on DictChain will print Int64 and not Int. Any good suggestions for this? Different tests depending on the machine?

ChrisRackauckas:
Just interpolate in the Int type. That fixed the test.

robertfeldt:
Ah, nicer solution than mine, thanks.

PR #69, opened 2017-07-06
Drop 0.4, fix 0.6 Julia support

alyst:
Fixes a number of Julia.Base v0.6 API deprecations.
Especially `X'` that returns `RowVector` and results in tests failures in v0.6.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,8 @@ os:
   - linux
   - osx
 julia:
-  - 0.4
   - 0.5
+  - 0.6
   - nightly
 matrix:
   allow_failures:
--- a/README.md
+++ b/README.md
@@ -7,9 +7,9 @@ BlackBoxOptim.jl
 
 [![Coverage Status](https://coveralls.io/repos/robertfeldt/BlackBoxOptim.jl/badge.png?branch=master)](https://coveralls.io/r/robertfeldt/BlackBoxOptim.jl?branch=master)
 
-[![BlackBoxOptim](http://pkg.julialang.org/badges/BlackBoxOptim_0.4.svg)](http://pkg.julialang.org/?pkg=BlackBoxOptim)
-
 [![BlackBoxOptim](http://pkg.julialang.org/badges/BlackBoxOptim_0.5.svg)](http://pkg.julialang.org/?pkg=BlackBoxOptim)
+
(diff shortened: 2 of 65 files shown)


alyst:
BTW, v0.3.0 of the package was tagged, but never published, so out-of-box version (i.e. `Pkg.add()`) doesn't work in Julia v0.6.
If it's ok to drop v0.4 Julia support at this stage, maybe it's worth to tag a new version (v0.4) upon merging this PR and publish it to Metadata.

robertfeldt:
Yes, thanks for doing this. I have partly done similar work on a branch. Will confirm and possibly merge our approaches this weekend and then tag and publish a new one to Metadata.

robertfeldt:
BTW, can some of your parallelization work be revisited now once 0.6 is out? If I remember correctly there were some language constructs you were waiting for?

alyst:
@robertfeldt I will shortly rebase my parallel population optimizer branch to the new master. The parallel evaluator is already there. Actually, I was waiting for was a threading support to see whether it would reduce the message passing overhead for the parallel evaluator. But threading is not there yet.

alyst:
I don't know whether Travis results are reliable indicators, but 0.6 runs take ~50% longer than 0.5.
Probably worth profiling.

robertfeldt:
It's disturbing that both from 0.4 to 0.5 and now from 0.5 to 0.6 performance have seemed to degrade. I guess Julia is too "lucid" a target yet...

robertfeldt:
This seems to be actually faster on 0.6 for me, at least for the standard test suite. Will benchmark some more before merging. Nice work. Any ideas how we should handle the deprecation warnings when running on 0.6? Just let them be for now?

alyst:
@robertfeldt The last two commits should handle typealias and abstract type deprecations. Fixing inner ctor deprecations would require either dropping 0.5 support because of the `where` keyword or using ugly conditional `eval("...")` constructs. But these remaining warnings are not so numerous. And now 0.6 looks faster for Travis too.

robertfeldt:
I tagged and published this as 0.2.2 for now. I need to update some things in the benchmark scripts before making this 0.3.0.

robertfeldt:
https://github.com/JuliaLang/METADATA.jl/pull/10346

PR #71, opened 2017-08-10
Add appveyor.yml

davidanthoff:
You'll also need to enable it over on https://www.appveyor.com/.

I'm seeing some test failures on Windows that I don't see in the travis logs, so would probably be good to routinely test on Windows via this.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,34 @@
+environment:
+  matrix:
+  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.5/julia-0.5-latest-win32.exe"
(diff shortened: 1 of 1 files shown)


robertfeldt:
In general this is something we want to do, yes. But you should know that testing stochastic algorithms is very hard and most of the time the failures found when running BBO.jl test suite is random variations since there is large variance when repeatedly running one and the same stochastic algorithm. So what you see is typically intermittent failures rather than "real" ones. But, since I have not tried it on Windows myself what you are seeing might be something else.

robertfeldt:
Can you please take a look at the failed checks and confirm they have nothing to do with your proposed changes in this PR @davidanthoff ? It seems they are unrelated.

davidanthoff:
Yes, this PR just adds one file that configures the appveyor CI testing, the travis failures must be due to something entirely different.

PR #80, opened 2018-04-15
drop Julia 0.5 and formatting cleanups

alyst:
With 0.7/1.0 on the horizon, this PR drops Julia 0.5 and switches to Julia 0.6 syntax for constructors (no implicit type parameters), method type parameters (`where`) and changes `type`/`immutable` into `[mutable] struct`.
So most of the warnings go away.
Also it fixes the docstring format (4-space indent for the function signature, no indent for the main body) and code indentation (4 spaces indents in all source files).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
   - linux
   - osx
 julia:
-  - 0.5
   - 0.6
   - nightly
 matrix:
--- a/README.md
+++ b/README.md
@@ -3,8 +3,8 @@ BlackBoxOptim.jl
 
 [![Build Status](https://travis-ci.org/robertfeldt/BlackBoxOptim.jl.svg?branch=master)](https://travis-ci.org/robertfeldt/BlackBoxOptim.jl)
 [![Coverage Status](https://coveralls.io/repos/robertfeldt/BlackBoxOptim.jl/badge.png?branch=master)](https://coveralls.io/r/robertfeldt/BlackBoxOptim.jl?branch=master)
-[![BlackBoxOptim](http://pkg.julialang.org/badges/BlackBoxOptim_0.4.svg)](http://pkg.julialang.org/?pkg=BlackBoxOptim)
-[![BlackBoxOptim](http://pkg.julialang.org/badges/BlackBoxOptim_0.5.svg)](http://pkg.julialang.org/?pkg=BlackBoxOptim)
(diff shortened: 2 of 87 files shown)


robertfeldt:
This looks good, thanks for doing this. Do you want to keep it open and work more on it or should we move to merge in the near-time?

alyst:
It should be more or less ready to go.
I'm using this branch internally at the moment to check whether there are any issues.
I'll let it distill for a few days and then remove the WIP tag.

alyst:
I've also fixed indentation/whitespace in most test files. Should be OK to merge.
I have a few enhancements on top of it that I'll submit once it's merged.

alyst:
should be ready to merge

robertfeldt:
This is unrelated to this PR but a minor thing I noted when running the benchmark is that it prints "DE modify state:" for each trace_state. git blame attributes the src code line as:

377e03ab (Alexey Stukalov 2016-04-14 23:36:57 +0200  32)         println(io, "DE modify state:")

in the differential_evolution.jl file. Any memories why this is there?

robertfeldt:
There clearly are performance regressions compared to the 2017-07-19 benchmark run but my analysis is that they are not related to this PR. I'm thus merging this in and will try to find the performance problems later. The deprecation warnings are very annoying and the changes you have introduced here overall increases code quality. Big thank you.

alyst:
@robertfeldt Thanks!

Re `trace_state()`: AFAIR I did it to check how the parameters of adaptive DE (cr/f) evolve over time. DE optimizer object (`DiffEvoOpt`) itself just applies the `modify` crossover operator, which implements the actual DE logic. So this `trace_state(de:DiffEvoOpt)` was supposed to call `trace_state(de.modify)`. It looks like there are 2 issues:
* I cannot locate the overloaded `trace_state(AdaptiveDiffEvoRandBin{N})` anymore, looks like it was lost in one of my WIP branches. So it defaults to the `trace_state()` of the generic genetic operator, which does nothing.
* The tracing condition `mode != :compact` doesn't seem right, it should be something like `mode == :verbose`. How can I reproduce the benchmark run where you observe this message?

robertfeldt:
If you go into `examples/benchmarking` and then run `julia runcompare.jl 2` you can see it. But it really is only to give TraceMode = :verbose that is the trick.

PR #83, opened 2018-05-17
Misc (mostly Borg) tweaks

alyst:
- Borg should use its own `MaxStepsWithoutEpsProgress` and not `MaxStepsWithoutProgress`, which defines the generic stopping criterion
- instead of throwing an error immediately if Pareto frontier grows too big, wait for a few iterations and stop (returning the intermediate result and specifying the proper stop reason) if it doesn't shrink back
- add `fitness_scheme_type(::Type{OptProblem})` and `fitness_type(::Type{OptProblem})` to aid in metaprogramming
- `TransformedProblem`: add `P` type parameter which is the type of the original problem; don't assume that the transformed problem keeps the same search space or fitness schema
- `BBO.isinterrupted(result)` to check whether the stop reason is interruption by user

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,15 +3,17 @@ os:
   - linux
   - osx
 julia:
-#  - 0.7 uncomment once 0.7 is out
+  - 0.7
+  - 1.0
   - nightly
-#matrix: uncomment once 0.7 is out
-#  allow_failures:
-#    - julia: nightly
+matrix:
+  allow_failures:
+    - julia: nightly
 notifications:
   email: false
-script:
-  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-  - julia --check-bounds=yes -e 'Pkg.clone(pwd()); Pkg.build("BlackBoxOptim"); Pkg.test("BlackBoxOptim"; coverage=true)'
+# uncomment the following lines to override the default test script
+#script:
+#  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
+#  - julia --check-bounds=yes -e 'using Pkg; Pkg.clone(pwd()); Pkg.build("RData"); Pkg.test("RData"; coverage=true)'
 after_success:
-  - julia -e 'cd(Pkg.dir("BlackBoxOptim")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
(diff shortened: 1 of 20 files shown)


robertfeldt:
This looks fine to me (thanks!) but I saw you added a commit yesterday. Is this ready to go or should we wait and do more minor tweaks?

alyst:
It should be ready for reivew.
I've enabled the tests for `ShiftedAndBiasedProblem` and updated the syntax in some other files in `tests/` that are not run, just for consistency.
But the code that is tested there is currently not used by BBO/benchmarks. So maybe we have to drop the stale code/tests at some point.

robertfeldt:
Hmm, again this looks nice but gives some performance regressions that I can't find the cause for. When running master on my MacBook `rake bench` takes around 1350 seconds (It's an old and not very powerful laptop ;)) while on this branch I've gotten values of 1499 to 1530 seconds. Most likely this is because of other fixes I did on master in between but I tried to merge and the problem persists. Strange.

alyst:
Hmm. Are there any specific tests that are being slowed down?
The change of `:MaxStepsWithoutEpsProgress` could have increased Borg running times because the stopping criterion is more stringent.
There's also switch to dot-broadcasting in `SearchSpace`, maybe it is slower than `map()`.

robertfeldt:
Unfortunately, I don't yet save the time taken by each run so I don't have that data. Let me extend the benchmarking code and try again. Unlikely to be the Borg-related code though since the benchmarking is still only done on single-objective problems...

alyst:
So .-broadcasting in RandomBounds indeed introduced a regression: all the `target` indices were unconditionally overwritten even when there were no change. With this reverted the times are improving.

[Benchmarking against the master](https://docs.google.com/spreadsheets/d/1_zH_ApQaMFJmzAtuKIY7ptATqAjy4JYt3AuaJnlyGwI/edit?usp=sharing) on my linux laptop using the updated benchmarking scripts (#91) shows no significant fitness regressions. (I was running it with the number of repetitions increased from 10 to 20, both for the master and the PR). There are, however, some significant elapsed time changes, but I don't see a pattern (other than 50/100-dim problems are more likely to have significant differences), these changes are in both directions. Probably it's some noise due to the fluctuations in the laptop CPU load (although I wasn't using it during benchmarking).

PR #86, opened 2018-07-13
added length method for DictChain

### LINKED ISSUE #85: No length method for DictChain
Trying to use JLD to save the results directly gives this error.

I'm making a pull request where I just define the length of a DictChain as the length of its vector. I'm assuming it was only omitted because it wasn't needed, but is there anything I'm missing?

grahamas:
fixes #85

### DIFF
--- a/src/parameters.jl
+++ b/src/parameters.jl
@@ -75,7 +75,72 @@ function Base.merge!(d::Dict{K,V}, dc::DictChain{K,V}) where {K,V}
     return d
 end
 
-Base.length(p::DictChain) = length(p.dicts)
+# Iterator methods
+function Base.start(dc::DictChain)
+    vec_state = start(dc.dicts)
+    return (vec_state, nothing)
+end
+
+function Base.next(dc::DictChain, state::Tuple{Int,Void})
+    vec_state = state[1]
+    (assoc, vec_state) = next(dc.dicts, vec_state)
+    assoc_state = start(assoc)
+    return next(dc, (vec_state, (assoc, assoc_state)))
+end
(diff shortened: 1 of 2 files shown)


robertfeldt:
Thanks for PR! But if we should see a DictChain as abstraction of a Dict that just happens ot be implemented as a series of dicts we might expect length to return the number of key-value pairs it contains. And then this implementation does not suffice. Does this provide as you expect for the JLD use case you got the error from? Can you provide a minimal code example of that JLD use case?

grahamas:
Fair point! Turns out there’s some discrepancy between JLD and JLD2. I was
using the latter; the former may also require iterator methods. I’ll figure
out what’s necessary and revise the PR, assuming I can get it to work as
the abstraction you’ve outlined (iterating through all elements of all
contained dicts).

grahamas:
Well, I implemented the natural iterator methods (returning successively all pairs in all dicts, allowing for nested chains). Unfortunately, JLD doesn't like that. It uses something to the effect of `(typeof(obj), [elt for elt in obj])` to serialize an object, and so the entire chain gets flattened and duplicate keys are overwritten. I don't think there's any way to deal with this without creating a JLD dependency? Anyway, JLD2 works fine even without this PR.

Do you still want these changes? I'll fix those errors I created.

robertfeldt:
Hmm, I don't use JLD myself so hard to judge. I'm leaning towards not making these changes and recommending people use JLD2 instead of JLD if they need this to work. Ok?

grahamas:
Works for me.

PR #91, opened 2018-08-10
Update benchmarking scripts

alyst:
While trying to investigate the cause of regressions in #83, I've updated the benchmarking scripts:
 * updated benchmark scripts to 0.7
 * added `ElapsedTime` comparison (it's an ugly copy-paste of the Fitness comparisons, but we anyway may consider rewriting benchmarking to use [BenchmarkTools.jl](https://github.com/JuliaCI/BenchmarkTools.jl))
 * added rake `benchref07` target for generating the reference data for benchmarks (basically calling `benchmarks/runupdate.jl`)
 * also added `bench07` target that calls `benchmarks/runcompare.jl` while using `BlackBoxOptim` `Project.toml` dependencies.

Unfortunately, running the benchmarks in 0.7 is not as straightforward as I have hoped. `Project/Manifest.toml` where supposed to bring much better control and automatization of the dependencies. It already does allow specifying custom dependencies for the benchmark (`[extras]`/`[targets]` in `Project.toml`), but currently the mechanisms to use these custom targets for code execution are not implemented (only the "test" target is supported by `Pkg.test()`). So all benchmarking dependencies would have to be specified in `[deps]`, which we most likely don't want to do. I omitted my .toml files from the commit (otherwise it will bloat BBO deps for all the users), but there's still `activate_pwd_env.jl`, which tells julia to use BBO's `Project.toml`. So some manual fixing of the local julia environment would be needed to run the benchmarks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,15 +3,17 @@ os:
   - linux
   - osx
 julia:
-#  - 0.7 uncomment once 0.7 is out
+  - 0.7
+  - 1.0
   - nightly
-#matrix: uncomment once 0.7 is out
-#  allow_failures:
-#    - julia: nightly
+matrix:
+  allow_failures:
+    - julia: nightly
 notifications:
   email: false
-script:
-  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-  - julia --check-bounds=yes -e 'Pkg.clone(pwd()); Pkg.build("BlackBoxOptim"); Pkg.test("BlackBoxOptim"; coverage=true)'
+# uncomment the following lines to override the default test script
+#script:
+#  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
+#  - julia --check-bounds=yes -e 'using Pkg; Pkg.clone(pwd()); Pkg.build("RData"); Pkg.test("RData"; coverage=true)'
 after_success:
-  - julia -e 'cd(Pkg.dir("BlackBoxOptim")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
+  - julia -e 'using Pkg, BlackBoxOptim; cd(joinpath(dirname(pathof(BlackBoxOptim)), "..")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(Coveralls.process_folder())'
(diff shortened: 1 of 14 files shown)


robertfeldt:
Thanks, this looks good. Yes, I think we need to rewrite to use BenchmarkTools or similar over time so this is temp for now. I don't fully understand the deps issues yet, are they why the build failed for 0.7 on Travis?

alyst:
>  I don't fully understand the deps issues yet, are they why the build failed for 0.7 on Travis?

No, because I haven't included .toml files to the PR (the old `REQUIRE` mechanism is still fully supported and I think we need to stick to it for a while).
Actually, the only failed target was v0.7 Linux, it looks like there was some network problem during `SpecialFunctions.jl` installation. So probably rerunning Travis would fix it.

As for a bit more details on the .toml issue...
Essentially, what we specify in `REQUIRE` is equivalent to `[deps]` section of `Project.toml`:

```toml
...
[deps]
Arpack = "7d9fca2a-8960-54d3-9f78-7d1dccf2cb97"
Distributions = "31c24e10-a181-5473-b8eb-7969acd0382f"
...
```

`test/REQUIRE` (test-only dependencies) are equivalent to listing the packages in `[extras]` section and then assigning them to "test" target in `[targets]` section:
```toml
...
[extras]
Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
...
[targets]
test = ["Test"]
...
```

Theoretically, it should also be possible to specify benchmark-only dependencies, e.g.
```toml
...
[extras]
ArgParse = "c7e460c6-2fb9-53a9-8c5b-16f535851c63"
CPUTime = "a9c8d775-2e2e-55fc-8582-045d282d599e"
Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
...
[targets]
test = ["Test", "CPUTime"]
bench = ["CPUTime", "ArgParse"]
...
```

But this is not implemented at the moment (there's no way to tell Julia to execute the code for "bench" target). So to make `rake bench` running from the package folder, I have locally put benchmark-only dependencies to `Project.toml` as normal dependencies (`[deps]` section). But we definitely don't want to merge this change and make BBO dependent on tons of packages.

robertfeldt:
Thanks, this looks fine. Benchmarking on my machine shows that on the smaller optimization problems running on 0.7 is slower while somewhat faster for longer-running problems. Going ahead though since I want to get these PRs merged before tagging a new release.

robertfeldt:
Just noting the benchmarking results on my machine, for the record:

```
0 significant fitness regressions at Benjamini-Hochberg 0.05 level

0 significant fitness improvments at Benjamini-Hochberg 0.05 level

82 significant time regressions at Benjamini-Hochberg 0.05 level

61 significant time improvments at Benjamini-Hochberg 0.05 level
elapsed time:    19 mins 3.5 secs
elapsed CPU time: 1128.760235 seconds
```

Seems clear that the move to 0.7 and the recent tweaks has slightly changed the timing. Overall my impression is that for longer running times and higher dimensions running on 0.7 is faster, while for shorter and/or lower-dimensional problems it is slower. This sounds like a decent trade-off.

Issue #82, opened 2018-05-11
Efficient Pareto frontier


alyst:
At the moment `EpsBoxArchive` type implementing Pareto frontier stores it as a simple `Vector{}`-based container with `O(M)` linear search.
Once the number of points (especially for N-objective problems) gets considerable (~5000), the overhead of maintaining the frontier (checking if candidate solution is dominated by the solutions from the frontier/removing frontier points dominated by the new candidate) becomes too high. Moreover, since the frontier is not parallelized, it becomes a real bottleneck for parallel BBO optimization.

For 2-objective problems we can solve the issue by sorting the solutions using the 1st component.
For N>2 objective problems we need something more advanced.
Without doing much research on the state-of-the-art containers for Pareto frontiers, the obvious candidate is [_k_-d tree](https://en.wikipedia.org/wiki/K-d_tree) (maybe it's even possible to use (N-1)-dimensional k-d trees as the frontier should be (N-1) manifold). [NearestNeighbours.jl](https://github.com/KristofferC/NearestNeighbors.jl) implements k-d trees, but these are the static ones; also I'm not sure whether it implements the range searches we need.

robertfeldt:
Yes, I've also been thinking about this. The Bkd-tree might be a good alternative:

https://klevas.mif.vu.lt/~algis/DSA/bkdtree.pdf

but I also have a half-ready implementation of Boundary Trees/Forest that might be useful:

http://www.aaai.org/ocs/index.php/AAAI/AAAI15/paper/download/9848/9953

I suggest we implement/release one or both of those as separate packages and then use it in BlackBoxOptim.

There are many alternatives though, some kind of location-sensitive hash might also be quite efficient.

robertfeldt:
Please tell me if you go ahead with any of these options so we don't overlap efforts. I think long-term, the multi-objective optimization can be a real "killer app" for BBO so would be nice to really work on these aspects as well as parallellization. Maybe we can aim to release these things quite quickly after Julia 1.0 is out?

robertfeldt:
Some relevant inspiration might be found here:

https://github.com/Keno/LSH.jl

but it seems it is not dynamic but requires the data upfront.

robertfeldt:
Seems his code is general enough that we could quite easily make a more dynamic version though. The functions and types to create the hash functions do not seem to require the data upfront. This is from a cursory look at his code only, though.

robertfeldt:
[...]

PR #107, opened 2018-11-22
use R*-Tree for Pareto frontier

alyst:
I've started [SpatialIndexing.jl](https://github.com/alyst/SpatialIndexing.jl) package (implements [R*-tree](https://en.wikipedia.org/wiki/R*_tree)), and this PR uses R*-tree as the container for the Pareto frontier in `EpsBoxArchive` as discussed in #82. Unfortunately, at the moment I'm hitting some internal Julia error JuliaLang/julia#30122. So either that would be fixed or I'll find some workaround to proceed further.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
   - linux
   - osx
 julia:
-  - 0.7
   - 1.0
   - nightly
 matrix:
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,6 +1,6 @@
-julia 0.7
+julia 1.0.3
 StatsBase
 Distributions
-SpatialIndexing
+SpatialIndexing 0.1
 Compat
 CPUTime
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,6 @@
 julia 0.7
 StatsBase
 Distributions
+SpatialIndexing
 Compat
 CPUTime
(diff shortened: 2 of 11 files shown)


robertfeldt:
Very nice. Please tell me when you feel it is stable enough to eval and merge.

alyst:
I was running it for a while with 24x parallel evaluator. It looks stable. It also reduces Pareto front update bottleneck, although that requires more benchmarking. JuliaLang/julia#30122 type intersection bug was fixed in 1.0.3/1.1.0/master, so for merging this PR we have to wait until any of these versions are released.

robertfeldt:
Ok, great, let's wait for at least 1.0.3 then. Did you need to do anything in particular for running he parallel evaluator or is this now well enough supported on master? Would be great if you can add some relevant examples/code for this just to help people get started.

alyst:
Julia 1.0.3 is released, so this PR should be fine for evaluation

> Did you need to do anything in particular for running he parallel evaluator or is this now well enough supported on master?

I was testing with async parallel evaluator from #46 (should be better for Borg). It works for me, but I'm hesitating a bit about about merging and promoting it, because it uses a strange/ugly mix of locks/semaphores and `SharedArrays` for synchronization (the result of trying to minimize the overhead of serialization during processes communication).

PR #109, opened 2018-12-21
Misc fixes

alyst:
Another round of fixes that seemed to me straightforward enough to put in a single PR, in particular:
 * parent-centric crossover implementation was quite inefficient (apparent for large problems), since it contained the computation of ndim X ndim temporary matrix; the updated implementation avoids it
 * updated `visualize_crossover.jl` visual test/example to use `Plots.jl` package
 * logging of optimization progress should be more readable (less digits after comma for performance metrics)
 * BBO's `rand_indexes()` method (which was a wrapper for `StatsBase` internal methods) replaced by `StatsBase.sample()` (since calls with keyword arguments should be properly optimized in 1.0)
 * `kwargs2dict()` should work in more cases now

### DIFF
--- a/examples/visualize_crossover.jl
+++ b/examples/visualize_crossover.jl
@@ -1,25 +1,25 @@
-using BlackBoxOptim, Gadfly
+using BlackBoxOptim, Plots, Random
 
 """
     plot_crossover(xover::CrossoverOperator, parents::Matrix{Float64};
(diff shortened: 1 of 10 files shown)


robertfeldt:
Thanks, this looks nice and I can't see a reason it would reduce coverage like that. Merging as is.

PR #111, opened 2019-01-12
SearchSpace: API cleanup + discrete dimensions support

alyst:
PR adds support for search spaces with discrete dimensions. I also used this opportunity to propose some cleanups to `SearchSpace` API.

**`SearchSpace` API changes**
  * `ContinuousSearchSpace` type was dropped -- it was not used, also it was ambiguous whether "continuous" refers to the functional-valued space or to finite-dimensional spaces with continuous-valued dimensions
  * `RangePerDimSearchSpace` renamed to `RectSearchSpace` (shorter + IMHO better describes the space)
  * `mins()`/`maxs()`/`deltas()` (aka `diameters()`) renamed to `dimmin()`/`dimmax()`/`dimdelta()` respectively. IMHO the new names better describe what these functions do and make them distinct from `Base.min/max`. Dropping the plural form allows `dimmin(ss, i)` methods to access individual dimensions (similar to `Base.size(arr, [i])`).
  * `ranges()`/`range_per_dim()` unified into `dimrange()`
  * `symmetric_search_space()` renamed to `RectSearchSpace`, since it's essentially an external constructor
  * `rand_individuals()`/`rand_individuals_lhs()` unified into `rand_individuals(...; method=)` with `:uniform` and `:latin_hypercube` (the default) methods currently supported.
  * `population()` also got `method=` keyword arg that is passed to `rand_individuals()`

The old types/functions still available via `@deprecate`/`@deprecate_binding` macros.

**Discrete dimensions**

The motivation is to better support discrete optimization problems + reduce the redundancy of populations for some continuous problems making the evolutionary optimization more efficient.
I have considered several other options, e.g.
* parameterize `RectSearchSpace{T}` and allow integer `T`, but that doesn't allow mixing discrete and continuous dimensions + may break some mutation/crossover operators
* add `discrete::Vector{Bool}` to `RectSearchSpace` that specifies the integer dimensions and automatically rounds them. However, when used to reduce redundancy, it may require rescaling the discrete dimensions back to the original range inside `fitness()` method, which introduces some overhead.

In this PR I've implemented the following:
* `RectSearchSpace` was renamed to `ContinuousRectSearchSpace`, and `RectSearchSpace` is its base abstract type
* added `MixedPrecisionsRectSearchSpace <: RectSearchSpace`. It has `dimdigits::Vector{Int}` field, which defines the digits of precision for each dimension (as in `Base.round(Float64, x, digits=)`. If `dimdigits[i]` < 0, `i` dimension is considered continuous. With this scheme, no rescaling of dimensions is required, while truly integral dimensions are also supported (`dimdigits[i]=0`).
* `RectSearchSpace()` constructor learned the new keyword argument `dimdigits` (defaults to `Nothing`). When no discrete dimensions are required, it constructs `ContinuousRectSearchSpace` as before.
* `RandomBound` embedding operator and `rand_individuals()` were updated to correctly project the individuals to `MixedPrecisionRectSearchSpace`.

### DIFF
--- a/examples/multiobjective.jl
+++ b/examples/multiobjective.jl
@@ -11,7 +11,7 @@ res = bboptimize(schaffer1; Method=:borg_moea,
 # N is the number of the problem (not fitness) dimensions
 pareto_curve_func(t, ::Type{Val{N}}) where {N} = (N*t[1]^2, N*(2-t[1])^2)
 pareto_curve = BlackBoxOptim.Hypersurface(pareto_curve_func,
-                                          symmetric_search_space(1, (0.0, 2.0)))
+                                          RectSearchSpace(1, (0.0, 2.0)))
 
 # generate the set of ϵ-indexed points on the exact Pareto frontier
 pareto_pts = BlackBoxOptim.generate(pareto_curve,
--- a/src/BlackBoxOptim.jl
+++ b/src/BlackBoxOptim.jl
@@ -62,7 +62,7 @@ export  Optimizer, AskTellOptimizer, SteppingOptimizer, PopulationOptimizer,
         ParamBounds, Individual, SearchSpace, FixedDimensionSearchSpace,
         RectSearchSpace, ContinuousRectSearchSpace,
         numdims, dimmin, dimmax, dimdelta, dimrange,
-        rand_individual, rand_individuals, rand_individuals_lhs,
+        rand_individual, rand_individuals,
 
         # Population
         FitPopulation,
--- a/src/BlackBoxOptim.jl
+++ b/src/BlackBoxOptim.jl
@@ -60,8 +60,8 @@ export  Optimizer, AskTellOptimizer, SteppingOptimizer, PopulationOptimizer,
 
         # Search spaces
         ParamBounds, Individual, SearchSpace, FixedDimensionSearchSpace,
-        RectSearchSpace, ContinuousRectSearchSpace,
-        numdims, dimmin, dimmax, dimdelta, dimrange,
+        RectSearchSpace, ContinuousRectSearchSpace, MixedPrecisionRectSearchSpace,
+        numdims, dimmin, dimmax, dimdelta, dimrange, dimdigits,
         rand_individual, rand_individuals,
 
         # Population
--- a/src/BlackBoxOptim.jl
+++ b/src/BlackBoxOptim.jl
@@ -61,7 +61,7 @@ export  Optimizer, AskTellOptimizer, SteppingOptimizer, PopulationOptimizer,
         # Search spaces
         ParamBounds, Individual, SearchSpace, FixedDimensionSearchSpace,
         RectSearchSpace, ContinuousRectSearchSpace, symmetric_search_space,
-        numdims, mins, maxs, deltas, ranges, range_for_dim, diameters,
+        numdims, dimmin, dimmax, dimdelta, dimrange,
         rand_individual, rand_individuals, rand_individuals_lhs,
 
         # Population
--- a/src/BlackBoxOptim.jl
+++ b/src/BlackBoxOptim.jl
@@ -60,7 +60,7 @@ export  Optimizer, AskTellOptimizer, SteppingOptimizer, PopulationOptimizer,
 
         # Search spaces
         ParamBounds, Individual, SearchSpace, FixedDimensionSearchSpace,
-        RectSearchSpace, ContinuousRectSearchSpace, symmetric_search_space,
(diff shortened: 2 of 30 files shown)


alyst:
I was modifying this PR to make sure the coverage doesn't regress, but now it should be ok and ready for review.

robertfeldt:
Thanks @alyst, I'm checking it now. Looking very nice and better support for discrete parameters would be a great feature. A few minor questions/comments:

- when I see `dim` I'm more thinking "dimensions" than "diameters" but I couldn't come up with better short forms (dimmin etc) so ok... `diammin` looks a bit strange.
- Not sure if `Base.in` should take the `dimdigits` into account. Currently it does not which should be ok for our use but maybe it will bite us if we forgot it also does not check it is a valid point in the space?
- What is the use case for `vcat` of search spaces?
- It makes sense that we leave any needed conversion of discrete parameters to Int for the user to handle in their fitness function. Maybe we should add an example with a mixed int and float vector being optimized that we can point to?

Rest looks great to me. Running performance tests now.

robertfeldt:
I run a comparison against the pre-PR master (d0f306a26526e81c4a94d03de1184bb24207542a):

Num problems with Time p-values < 0.05: 67
  Time is better in current run in 32 tests
  Time is worse in current run in 35 tests
Num problems with Time p-values < 0.01: 44
  Time is better in current run in 17 tests
  Time is worse in current run in 27 tests
0 significant fitness regressions at Benjamini-Hochberg 0.05 level
0 significant fitness improvements at Benjamini-Hochberg 0.05 level
31 significant time regressions at Benjamini-Hochberg 0.05 level
22 significant time improvements at Benjamini-Hochberg 0.05 level
elapsed time:    17 mins 38.4 secs
elapsed CPU time: 1056.340682 seconds

so it seems there are no major performance problems introduced by this PR.

alyst:
@robertfeldt Thanks for the review and for the benchmarks! (BTW, do you have a detailed table?)

> when I see dim I'm more thinking "dimensions" than "diameters" but I couldn't come up with better short forms (dimmin etc) so ok... diammin looks a bit strange.

That was the idea: `dimmin` = "**dim**ension **min**imum" etc. I was also thinking about names inspired by MBR (minimal bounding rectangle), e.g. `mbrlow` == `dimmin`, but I didn't like it that much. I'm totally open to any other alternatives; the intention is to have one method that returns either a vector with metrics for all dimensions or just a single metric for a specific dimension (which is useful when getting this metric is expensive).

> Not sure if Base.in should take the dimdigits into account. Currently it does not which should be ok for our use but maybe it will bite us if we forgot it also does not check it is a valid point in the space?

I think currently `in()` is used just in the tests. We can also use it to check the correctness of starting points/populations: it could be that user-specified `fitness(params)` will fail with rather cryptic `InexactError` during the conversion to `Int` if some `params[i]` have incorrect precision.

> What is the use case for vcat of search spaces?

To compose problems. We are using it in `FunctionBasedProblemFamily`: its first dimensions are "hardcoded" (`reserved_ss`), while the number of the remaining dimensions (all having the same range) is flexible; so we `vcat` the reserved search space and the "symmetric" one.

> It makes sense that we leave any needed conversion of discrete parameters to Int for the user to handle in their fitness function. Maybe we should add an example with a mixed int and float vector being optimized that we can point to?

An example mixed problem is a great idea. Do you have some nice problem in mind?

robertfeldt:
Ok, let's go with `dimmin` et al. I think naming is very hard, ;), and I can't see a clearly better alternative.

Let's keep Base.in for now. It does make sense in the tests. And one can argue that even "invalid" points that are inside the Rect borders are actually "in" the space.

`vcat` makes sense, I had forgot that we actually used it and though you just added it for "completeness" which didn't make sense to me.

I did do a mixed problem just to try things out on the branch but I think it might be too complex to really show this (it is a kind of "genetic programming" where the selection of an operation is via an int and it's parameter(s) are floats so the optimized vector is a mix of "ints" and floats). But the complexity of it probably "hides" what it is trying to exemplify so not sure it is a good one. I'll possibly put it in a "spikes" directory after we have merged this PR and we can discuss it more then.

For the performance benchmarking I really need to clean up the scripts there so they are more useful and give better/shorter reports on the perf comparisons.

Cheers, Robert

alyst:
Thanks! Do you think it's the right moment to tag v0.5.0?
Since v0.4.0 in August:
* SpatialIndexing-powered Pareto front
* SearchSpace API changes
* discrete dimensions
* better 1.0 compatibility
* multiple fixes

PR #118, opened 2019-01-29
Add Julia v1.1 to CI tests

alyst:


alyst:
For some reason I stopped seeing "Some checks didn't complete yet" yellowish status line for some time, although when the CI tests finish the status gets updated.
I'm not an expert in CI configs, but maybe there are some options on GitHub/Travis/AppVeyor side that could be updated.

PR #124, opened 2019-06-13
Create ODEwithFixParams.jl

ayadlin:


ayadlin:
The file is an example of optimization for DAE parameters while leaving some fixed.
Sorry i forgot to comment before

PR #130, opened 2019-07-26
Correct optimizer name in README

karajan9:
I think (and according to my testing how this works) this should be the correct name for the optimizer.


robertfeldt:
Thanks!

PR #150, opened 2020-05-04
noting some of the high-level goals with redesign and including examples

robertfeldt:


robertfeldt:
This should not yet be a pull request... My bad :)

PR #156, opened 2020-09-30
Tolerate more kinds of bounds

baggepinnen:
This allows, e.g., `[(1, 2.5)]` which is a mix of ints and floats.


robertfeldt:
Thanks!

PR #159, opened 2020-10-21
more generous type allowances

baggepinnen:
This is one more PR making the optimizer accept any form of vector with any form of valid tuples specifying the search space.


robertfeldt:
Thanks, Fredrik!

PR #161, opened 2020-12-14
Add CompatHelper

devmotion:
I noticed that BlackBoxOptim is not compatible with Distributions 0.24. The problem is that Project.toml does not explicitly state compatibilities of the dependencies and does not upper bound them, and therefore they were upper bounded automatically in the registry: https://github.com/JuliaRegistries/General/blob/master/B/BlackBoxOptim/Compat.toml

This PR adds CompatHelper which should help you to keep Project.toml up to date. Moreover, it updates TagBot.

### DIFF
--- a/.github/workflows/CompatHelper.yml
+++ b/.github/workflows/CompatHelper.yml
@@ -1,0 +1,15 @@
+name: CompatHelper
+on:
+  schedule:
+    - cron: '00 00 * * *'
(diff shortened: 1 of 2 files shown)


ChrisRackauckas:
Bump

robertfeldt:
Thanks, seems useful.

PR #162, opened 2021-02-02
Add a docs breadcrumb for supported kwargs

mbauman:
Not sure if this is how you'd prefer this to be documented (or if it's even always accurate), but I had a hard time discovering this and figured the breadcrumb would be helpful.


robertfeldt:
Thanks!

Issue #163, opened 2021-02-12
Int overflow in ϵ_index


Libbum:
So I thought I've been hitting the problem in #78 for a while, but it turns out not to be the case. I clamped all of my outputs and fitness function to [0, 1e8] and was still having problems.

[This floor call](https://github.com/Libbum/BlackBoxOptim.jl/blob/65a0c207d8c5cfaf8e298e311f9b2cdf9501e18f/src/ntuple_fitness.jl#L136) is actually the problem.
```julia
@inline function ϵ_index(u::F, ϵ::F, ::Type{Val{true}}) where F
    if isnan(u)
        return (typemax(Int), zero(F))
    else
        u_div_ϵ = clamp(u/ϵ, convert(F, typemin(Int)), convert(F, typemax(Int)))
        ix = floor(Int, u_div_ϵ+10eps(F))
        return (ix, max(zero(F), u_div_ϵ-ix))
    end
end
```

When clamping `u_div_ϵ`, there's the possibility you'll get `u_div_ϵ = convert(F, typemax(Int))`. On the next line, adding the `eps()` certainly doesn't help, but even so: `floor(Int, convert(F,typemax(Int)))` is going to fail since the float is larger than the safe float conversion value. 

```julia
julia> maxintfloat(Float64)
9.007199254740992e15

julia> convert(Float64,typemax(Int))
9.223372036854776e18
```

My fix for this is too ugly, hence this issue rather than a PR. I'm just not versed enough in these conversion problems.
```julia
ix = convert(Int, min(floor(Int128, u_div_ϵ+10eps(F)), typemax(Int)))
```
In other words: allow the float to spill over into `typemax(Int64)+1`, then artificially bring it back into an`Int64` range. Seems like a bit of a hack to me...

DrDJIng:
This issue happens using floor in the minimizing scheme, but I see the maximizing scheme uses ceil in a [similar implementation](https://github.com/Libbum/BlackBoxOptim.jl/blob/65a0c207d8c5cfaf8e298e311f9b2cdf9501e18f/src/ntuple_fitness.jl#L147). 
Do you get this issue using ceil as well?

Libbum:
Yes, that would be the case. So it looks like both of the inline ϵ_index functions are in need of attention in that regard.

robertfeldt:
[...]

PR #164, opened 2021-02-13
Safely truncate large floats

Libbum:
An attempt at solving #163. Unsure if it's in the style you like, but satisfies the edge case. 

Ref: https://discourse.julialang.org/t/truncating-floats-into-a-specific-int-size/55207


robertfeldt:
Ok, I've been thinking about this one and since I don't expect people to run into these kinds of problems very frequently it doesn't seems called for implementing different fitness schemes for the different behaviors one might want here. Thus, I think it is ok to merge your PR but I will then add a warning message to the (hopefully rare) situation when the clamping is needed.

PR #178, opened 2021-04-28
format docu of OptRunController as list

bgctw:
Thanks for this useful package.
I am learning about the available options and discovered the link to the docu of `OptRunController`. However, this docu is currently not formatted properly in the Julia help, because of spaces at the beginnings of the lines and hence difficult to read. Therefore I create this pull request. Please, do not mind the few other formatting  long comments in bboptimize.jl.


robertfeldt:
Great, thanks!

PR #192, opened 2022-01-27
Providing multiple initial candidates to multi-objective optimization

sd109:
This branch adds the option to specify multiple 'x0' candidates within the starting population for the borg-moea algorithm. 

The current implementation works and I have been using it myself, but may not be the cleanest way to do things. If this feature is something that you're interested in merging then I'd be happy think about tidying up the implementation if needed.

### DIFF
--- a/src/bboptimize.jl
+++ b/src/bboptimize.jl
@@ -65,8 +65,10 @@ function bboptimize(optctrl::OptController, x0 = nothing; kwargs...)
     if length(kwargs) > 0
         update_parameters!(optctrl, kwargs2dict(kwargs))
     end
-    if !isnothing(x0)
+    if isa(x0, Vector{T} where T <: Number) #Provided a single x0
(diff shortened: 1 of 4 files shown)


robertfeldt:
Thanks for this, it looks useful. Maybe the type should be more restrictive (Vector{Vector{<:Number}})?

robertfeldt:
Also this should probably be allowed for all PopulationOptimizers not only BorgMOEA.

robertfeldt:
Thanks. I decided to do this slightly differently (since I wanted to depend on Random rather than StatsBase) and also implement for all the population-based optimisers. I haven't tested with Borg MOEA yet though so feel free to try it and please report back if doesn't work. See this commit:

https://github.com/robertfeldt/BlackBoxOptim.jl/commit/2ddf6e216bd836de9374d7725b6f7c94350e70d4

PR #198, opened 2022-05-24
Fix update_fitness!() for MultithreadEvaluator.

alyst:
`update_fitness!()` is expected to return the same vector of candidates it has been given, but for *MultithreadEvaluator* it was sometimes returning
*nothing* if there was no work to do. With the PR, it always returns the candidates vector.
That fixes *rank_by_fitness!()* that relies on *update_fitness!()* return type.

### DIFF
--- a/src/multithread_evaluator.jl
+++ b/src/multithread_evaluator.jl
@@ -416,8 +416,6 @@ function async_update_fitness!(eval::MultithreadEvaluator, candidates::Any; forc
     return job
 end
 
-sync_update_fitness(f::Any, job::Nothing, eval::MultithreadEvaluator) = nothing
-
 # Waits until MTFitnessEvalJob is fully processed.
 # Constantly monitors job.results for the new candidates and adds them
 # to the fitnesses archive.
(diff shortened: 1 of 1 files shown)


robertfeldt:
Thanks, looks good.

What is the status of the MultithreadEvaluator now? Are you actively using it? There has been reports people have troubles running BBO on multiple threads and I've considered going for a simpler solution that just Threads for loops over the fitness evaluation calls. Any advice?

alyst:
> What is the status of the MultithreadEvaluator now? Are you actively using it? There has been reports people have troubles running BBO on multiple threads and I've considered going for a simpler solution that just Threads for loops over the fitness evaluation calls. Any advice?

Unfortunately, I don't have time to invest into improving *MultithreadEvaluator*, this was just a small fix.
I expect that in the recent Julia version there might be better ways to control the threads, which may improve the performance of *MultithreadEvaluator* (AFAIR controlling threads was the main headache when implementing it).
I think `@threads for candi in candidates` should work fine if there are at dozens of candidates or the fitness is very expensive, otherwise, I'm afraid, the overhead of managing threads (behind the scenes, `@thread for` creates and schedules parallel tasks). Actually, the only current BBO algorithm that should benefit from multithread evaluation is Borg, because it allows scheduling fitness calculation for multiple candidates plus its main thread can do some other work, while fitness is being calculated. For the other algorithms *MultithreadEvaluator* would work, but since there's typically no more than 1-2 synchronous fitness evaluations at a time, there's no performance benefit. But I guess it is possible to modify existing algorithms for parallel execution (e.g. simultaneous mutations & recombinations).

robertfeldt:
Ok, great input thanks.

Yes, I have started on a branch that will do multiple calls to existing algs to collect a batch of candidates which can then be evaluated in parallel (threads). At least for population-based algorithms this should be fine while for generating set search and other in the hillclimbing style only have one current best which makes it less clear how to best use multiple threads. OTOH, even GSS internally creates a batch of "directions" with which to change the current best so it should be possible to eval them in a batch.

PR #216, opened 2023-06-17
Refactor real-time plots, make HTTP.jl optional

alyst:
This PR refactors *VegaLiteGraphFitnessGraph* into plotting-engine-agnostic *RealtimePlot* and moves *VegaLite* and *HTTP*-specific code into *ext/BlackBoxOptimRelatimePlotServerExt.jl* (uses package extension mechanism on Julia 1.9, or *Requires.jl* on earlier version).
So *HTTP.jl* and all its dependencies become optional, which reduces overhead for projects that use *BlackBoxOptim.jl* only for computation.

Also
closes #214


robertfeldt (commented):
Looks good, thanks.

robertfeldt (approved):
Approved

alyst:
@robertfeldt  Thank you, that was very fast! I've already updated the version in `Project.toml` to 0.6.3, could you please tag&release the new version?

Issue #78, opened 2018-03-06
Inexact Error() when using BorgMOEA


mcfefa:
I'm getting an InexactError() when I try to use the BorgMOEA method to fit multiple objectives at once. Attached is a Jupyter Notebook with a toy ODE system that I'm trying to fit to some experimental data where I have multiple initial conditions (which I set up with a Monte Carlo Problem). 

[BorgMOEA-Test.zip](https://github.com/robertfeldt/BlackBoxOptim.jl/files/1786704/BorgMOEA-Test.zip)

It seems if I have a smaller population size I either won't get the error or the optimization will run a few times before throwing the error.

Is this a known bug, or user error?

Output statement: 

Starting optimization with optimizer BlackBoxOptim.BorgMOEA{BlackBoxOptim.EpsBoxDominanceFitnessScheme{3,Float64,true,Base.#sum},BlackBoxOptim.ProblemEvaluator{Tuple{Float64,Float64,Float64},BlackBoxOptim.IndexedTupleFitness{3,Float64},BlackBoxOptim.EpsBoxArchive{3,Float64,BlackBoxOptim.EpsBoxDominanceFitnessScheme{3,Float64,true,Base.#sum}},BlackBoxOptim.FunctionBasedProblem{BlackBoxOptim.ParetoFitnessScheme{3,Float64,true,Base.#sum},BlackBoxOptim.RangePerDimSearchSpace,Void}},BlackBoxOptim.FitPopulation{BlackBoxOptim.IndexedTupleFitness{3,Float64}},BlackBoxOptim.FixedGeneticOperatorsMixture,BlackBoxOptim.RandomBound{BlackBoxOptim.RangePerDimSearchSpace}}
0.00 secs, 0 evals, 0 steps
pop.size=1000 arch.size=0 n.restarts=0
InexactError()

Stacktrace:
 [1] trunc(::Type{Int64}, ::Float64) at ./float.jl:672
 [2] ϵ_index at /Users/mcfefa/.julia/v0.6/BlackBoxOptim/src/ntuple_fitness.jl:128 [inlined]
 [3] macro expansion at /Users/mcfefa/.julia/v0.6/BlackBoxOptim/src/ntuple_fitness.jl:147 [inlined]
[...]

PR #217, opened 2023-06-18
Switch to GitHub CI 

alyst:
Travis/AppVeyor CI Julia infrastructure is aging, this PR switches to Julia github workflow.

**Upd**: when checking the tests, I've discovered that eps-box fitness tests fail on x86, so I had to refactor how we are clamping fitness into integer range. I hope the fix would not trigger *InexactError* again. Ref #78, #163.


robertfeldt (approved):
nice thx

PR #237, opened 2024-12-09
slightly simplify type definitions: `Tuple{<:T}` -> `Tuple{T}`

nsajko:
```julia-repl
julia> Tuple{<:Real, <:Real} == Tuple{Real, Real}
true
```


robertfeldt (approved):
Looks good.

PR #279, opened 2026-07-20
Require HTTP 1 for optional integration

ChrisRackauckas-Claude:
## Summary

Ignore this PR until it has been reviewed by @ChrisRackauckas.

- require HTTP 1 or 2.1+ for the optional realtime-plot dependency
- retire HTTP 0.9, which has been superseded by stable 1.x since 2021

HTTP 2 is still recent, so this keeps HTTP 1 during the transition. The change is intentionally limited to the shared weak/test dependency compat entry.

## Local validation

- Exact all-dependency downgrade resolution on Julia 1.10: selected HTTP 1.0.0.
- Locked exact-floor `GROUP=Core` on Julia 1.10: passed.
- Exact optional-dependency load probe: BlackBoxOptim 0.6.9 and HTTP 1.0.0 co-loaded successfully.
- Current-resolution `GROUP=Core` on Julia 1.12: passed.
- Current-resolution `GROUP=QA` on Julia 1.12: passed (14 passed, 5 pre-existing broken checks).
- `Runic.main(["--check", "src", "test"])`: passed. The repository-wide check fails identically on clean upstream because legacy `spikes/` and benchmarking files do not parse as current Julia.
- `git diff --check`: passed.


ChrisRackauckas-Claude:
Scratchpad update for commit `e5459d6`:

- Kept the change to the single HTTP compat line.
- Verified the exact all-dependency graph selects HTTP 1.0.0.
- Ran the locked Julia 1.10 Core suite and an isolated BlackBoxOptim 0.6.9 + HTTP 1.0.0 co-load probe successfully.
- Ran current Julia 1.12 Core and QA successfully (QA: 14 passed, 5 pre-existing broken checks).
- Ran Runic across the repository; it fails on the same legacy unparsable `spikes/` and benchmarking files on clean upstream. The maintained `src/` and `test/` trees pass Runic, and the two full-check logs are byte-identical.
- `git diff --check` passes.

No additional code changes are planned; waiting for CI review.

